PR #2, opened 2020-07-21
feat: add build documentation

matthew-chirgwin:
- Offers build documentation, with code to follow in another PR
- Covers the approach, tools used, the configuration, and end output for UI code
- A follow up PR will add (once agreed) how the built UI code is then built/deployed into the wider Strimzi

### DIFF
--- a/docs/Build.md
+++ b/docs/Build.md
@@ -71,10 +71,10 @@ Webpack allows file specific loaders or utilities to be invoked as a part of the
 
 | Rule        | Plugin/loader(s)           | Purpose  |
 | ------------- | ------------- | -----:|
-| `/(\.css|.scss)$/`      | `style-loader` (dev only), `miniCssExtractPlugin.loader` (production only), `css-loader`, `sass-loader` | Handle scss/css loading/references. If dev mode, use `style-loader` for speed, else use `miniCssExtractPlugin.loader` (in combination with the `miniCssExtractPlugin` plugin) to produce/emit a css file |
+| `/(\.css\|.scss)$/`      | `style-loader` (dev only), `miniCssExtractPlugin.loader` (production only), `css-loader`, `sass-loader` | Handle scss/css loading/references. If dev mode, use `style-loader` for speed, else use `miniCssExtractPlugin.loader` (in combination with the `miniCssExtractPlugin` plugin) to produce/emit a css file |
 | `/(\.js)$/`      | `babel-loader` | Perform babel transpile on all JS files |
-| `/\.(woff(2)?|ttf|eot)$/`      | `file-loader` | For any font file, use file-loader to package the font to the `output.path` and replace/update any imports of those fonts to this location. These will be directed to a 'fonts' directory. |
-| `/\.(jpg|gif|png|svg)$/`      | `file-loader` | For any image file, use file-loader to package the image to the `output.path` and replace/update any imports of those images to this location. These will be directed to a 'images' directory. |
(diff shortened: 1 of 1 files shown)


samuel-hawker:
This looks good to me! Well thought out and all of the choices make sense to me, though I wouldn't want to approve until some of the other working developers on the UI approve saying they like all the choices you have made here. :)

matthew-chirgwin:
> This looks good to me! Well thought out and all of the choices make sense to me, though I wouldn't want to approve until some of the other working developers on the UI approve saying they like all the choices you have made here. :)

Thanks @samuel-hawker - I have requested a review from them already :)

docs/Build.md
```diff
@@ -0,0 +1,145 @@
...
+
+For convenience, all top level module directories will be automatically aliased via a helper in the webpack configuration, as well defining a view alias.
+
+#### Webpack configuration and plugins
+
+In addition to aliasing, the webpack configuration will be as follows:
+
+| Option        | Value           | Purpose  |
+| ------------- | ------------- | -----:|
+| entry      | `client/Bootstrap/index.js` | Build entry point. Note the `scss/css` is expected to be imported by individual view layer code, which is imported (directly or indirectly) from this file  |
```
nictownsend: Casual mention over scss/css imports - is that (to be) mentioned elsewhere?

nictownsend (approved):

docs/Build.md
matthew-chirgwin: I was going to be explicit in a 'Code Style' doc PR, but my suggestion would be a (s)css import per component (as a part of a given view layer). The CSS optimisation plugin will then deduplicate down to a single set of class names. This should keep builds fast when developing (ie just one file changing) but also minimal when built in production. I will update the doc to mention details to follow - plus a few other small tweaks.

nictownsend (approved):

scholzj (commented):
I don't really understand much about the content. But ping me when you think it i sready - I can approve and merge it ;-)

docs/Build.md
nictownsend: Will optimisation namespace where duplicates happen? Or fail build (hopefully this)?

seanforyou23 (approved):
Nice work @matthew-chirgwin! This looks great 😎

docs/Build.md
matthew-chirgwin: OOTB, it will merge:
File one:

```
.foo {
color: red;
height: 100%;
}
```
File two:
```
.foo {
color: green;
}
.bar: {
color: red;
}
```

will become:

```
.foo {
color: green;
height: 100%;
}
.bar: {
color: red;
}
```

when built, with standard css precedence (ie later takes precedence) based on import order.

If it is ok, I will look at options here, and if we want to fail on conflicts - thoughts welcome. I can see pro's and cons either way. Based on discussion, I will update the documentation here with our decision in a follow up PR. Worth noting this optimisation only applies on production builds - when in development, we would see multiple (identical) class names (say), and where they originate from.

nictownsend: Yep sounds good. My main concern is over whether we have to enforce name-spacing as a code style, or it can be explicitly forced.

matthew-chirgwin:
Thank you for reviews @nictownsend and @seanforyou23 ! I think we are all good to merge now @scholzj - a PR to implement the above will follow shortly. @nictownsend / @seanforyou23 would you mind reviewing that when it is raised?

ppatierno (approved):
Everything seems to be fine as @matthew-chirgwin said :-) so I am happy to approve!

scholzj (approved):

scholzj:
@samuel-hawker Do you have anything more? Or can I merge it? Thanks.

samuel-hawker:
Nothing more to add from me, this is a great start!

samuel-hawker (approved):

PR #18, opened 2020-08-28
feat: add documentation for navigation capabilities in the UI

### LINKED ISSUE #12: Architecture: routing, navigation, introspection

Investigate and propose an approach for routing and navigation, and how GQL introspection can dynamically enable/disable routes in the UI

matthew-chirgwin:
- Adds documentation describing how introspection, navigation and routing will work in the Strimzi-ui

Closes: #12


scholzj (approved):

dlabaj (approved):

client/Contexts/Introspect/README.md
```diff
@@ -0,0 +1,79 @@
...
+A GraphQL schema is made up of `queries`, `types`, `mutations` and `subscriptions`. Together they describe what the server hosts and is capable of, but introspection is required to understand the relationship between these constructs, and what `mutation` for example will update a particular `type`.
+
+To drive this introspection process, the Strimzi ui defines what it requires/expects to operate at an entity level. This approach will define what GraphQL `types` are expected, the fields those types should contain (and what types they are), the operations (`mutations`) that can be performed on a given `type`, and what `subscriptions` are available for a given type as well. Using the example of a Topic, this would look as follows:
+
+```
+{
+    Topic: {
+        type: 'Topic', // the GraphQL type name
+        fields: {
+            name: GQL_STRING, // the expected GraphQL type - could be string or object - which is checked if is available and thus query-able
```
nictownsend, line 16: If we have an "unreliable" schema (hence the introspection) - would it be simpler to have a translation layer?

As in - components call `buildQuery` with the "complete" query, and fields are filtered back out? And the same way with a `parseResults`, it results in an object with `null`s for bits that are missing?

My concern is that the components now have to build composed queries based off fields/capabilities.

What would be great is if introspection told the component what "capabilities" can be used (and those control features) - but the data in/out was just filtered out so that you just handled null if something you needed was missing?

client/Bootstrap/Navigation/README.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Example usage
+
+To follow in a later PR.
+
+_Note/To do_: Describe/define how path parameters, as well as helper functions, such as 'go back to the previous Strimzi ui page', 'go to this page by ID', or 'is this page available by ID', are passed to pages mounted by the router.
+
+## Route configuration processing
+
+As mentioned, this component takes in [configuration](../../Pages/README.md) which maps [`Panels`](../../Panels/README.md) to routes, along with other supporting metadata, such as a translatable page names, what backend services are required to show the page, and what authorization(s) a user may need to see it. This configuration is processed via the Navigation's model, and will result in an object which is then iterated on in the Navigation's view layers to generate the UI's routing/navigation. For example, given the following configuration:
```
nictownsend, line 61: `home/topic/1/partitions` - what if that was to render the topic page, but with a slide out panel enabled?

Might be me thinking too hard about this - but it looks like this is tying us such that a path maps to a single component.

And is there a way to add dynamic properties (i.e pass through something like a query param or a route param) - so that the component is unaware of the history or route objects?

client/Pages/README.md
```diff
@@ -0,0 +1,103 @@
+# Pages
+
+This module defines and exports metadata which describes the pages in the Strimzi UI, and how they relate to the [`Panel`](../Panels/README.md) components that implement them. This metadata is then processed by the UI's [navigation logic](../Bootstrap/Navigation) at runtime to determine what pages are shown and available to a user, but also how they are rendered in context to the wider UI and relate to one another. The motivation for this is three fold: to separate the components that implement a page and navigation (to avoid cross contamination), to enable extensibility for new pages in future, and allow re use of the same `Panel` components in numerous contexts with minimal changes.
+
+## Page Metadata Schema
+
+The schema used to define pages in the Strimzi-ui is designed around capability, and the scenarios those capabilities are presented in. The exported object should be named to reflect the capability it provides, and the metadata contained to specify the `Panel` component used to provide it (`contentComponent`), and where (and how) it should be presented (`contexts`). The below example (pseudo) shape shows the structure expected:
+
+```
```
nictownsend, line 9: Is this a duplication of the content above?

client/Bootstrap/Navigation/README.md
```diff
@@ -0,0 +1,271 @@
...
+                feature_flag: 'PAGE.TOPIC.CONSUMER_GROUPS',
+                order: 0,
+                icon: undefined,
+                pageType: NORMAL,
+                properties: {
+                    mode: 'Topic'
+                },
+                requiresMinimum: {
+                    backendSupportFor: {
+                        Group: { READ }
```
nictownsend, line 117: Where are these constants defined? And what about subscription support?

client/Bootstrap/Navigation/README.md
matthew-chirgwin: For elements like that, I would consider a query parameter perhaps. The job of the Navigation is to render routes based on configuration, and pass contextual information (path parameters, query parameters) to the configured component. What the component does with those pieces of information are not in my opinion the responsibility of the navigation component.

To your second point, I have prototyped and validated the ability to pass things like path paramters, query parameters via a property to the component on render. I did leave it as a todo for implementation, but I will update the PR with more details on this now.

client/Bootstrap/Navigation/README.md
matthew-chirgwin: Did you see https://github.com/matthew-chirgwin/strimzi-ui/blob/12-routing-nav/docs/Architecture.md#entity-model ? I do not mention where the common constants come from, so will add that, but the idea is that the 'abilities' on the entity could be anything, and that they are contextual, known and used/referenced by the code that needs them.

client/Pages/README.md
matthew-chirgwin: By above, do you mean in the Pages README? If so, I see the text above this section as more of a summary, and this as more detail about the structure itself.

client/Contexts/Introspect/README.md
matthew-chirgwin: > What would be great is if introspection told the component what "capabilities" can be used (and those control features)

That is the intent here - this model descirbes what is expected, which can then be checked with flags being enabled /disabled for example. The only usage so far is to show/hide whole pages, but this could be used at a feature flag level as well.

I must admit, I have not considered a partial schema case (and returning nulls/empty values) for a query. Let me look into this and see what would need to change to support it.

matthew-chirgwin:
@dlabaj / @nictownsend / @scholzj thank you all for the reviews/comments. I have just pushed an update to address them. Please let me know if you have any further thoughts.

@nictownsend re 'filling in blanks' around queries - I think the introspection result could be used to help guide a query helper (or similar) to fill in blanks if required. I do err however that it may then cause other issues (if users of the query result do not handle nulls etc well). Would like to discuss with you further on it though.

dlabaj (approved):

nictownsend (changes requested):
LGTM except for one small point re `backendSupportFor` - how can you articulate the need for a subscription vs query?

matthew-chirgwin:
> LGTM except for one small point re `backendSupportFor` - how can you articulate the need for a subscription vs query?

Where a subscription is required, this would be an additional `backendSupportFor` key to check for it. At which point the callback for that key would be proivided the schema. There is an example of this here: https://github.com/matthew-chirgwin/strimzi-ui/blob/12-routing-nav/docs/Architecture.md#entity-model (`liveUpdates`)

nictownsend (approved):
Thanks, LGTM for laying groundwork

PR #22, opened 2020-09-23
feat: Add test documentation

### LINKED ISSUE #7: Provide UI Test documentation
Issue to write up and provide test documentation for Strimzi-ui.

This should cover:

- The types of testing required on areas of the codebase
  - I would suggest we aim for 100% code coverage as a minimum from day 0 - which should be checked/achieved by running tests on just the area in question
- Tools used for testing and why
  - Given piror discussion, would suggest Jest as a test runner/framework, React testing library
- How accesibility is tested, and accessible status maintained
- Updating existing documentation to show how/why these tools/approaches are used/how and where things should be tested

Initially this will cover the UI codebase only, but once we integrate the UI with Strimzi, it should also cover how these integrations are tested (eg via system test)

JQrdan:
- Adds test overview documentation
- Adds documentation for test specific files and directories

Closes: #7

### DIFF
--- a/client/Bootstrap/README.md
+++ b/client/Bootstrap/README.md
@@ -6,7 +6,11 @@ While it's own directory, the code in this directory should be tested and develo
 
 ## Test approach
 
-Bootstrap should be tested in a behavioural manor. See [Behavioural Driven Development](../../docs/Test.md#style-of-test).
+The code in this directory should only be tested through End-to-End tests.
+This is because these files do not contribute unit-level behaviour but have
+an impact on the entire UI. This should help speed up development of bootstrap
+code and unit testing at this level can be fiddly. See
+[End-to-End testing](../../docs/Test.md#end-to-end-testing).
 
 ## Expected files
 
--- a/client/Bootstrap/README.md
+++ b/client/Bootstrap/README.md
@@ -6,7 +6,7 @@ While it's own directory, the code in this directory should be tested and develo
 
 ## Test approach
 
-Bootstrap should be tested in a behavioural manor. See [Behavioural test style](../../docs/Test.md#behavioural-test-style).
+Bootstrap should be tested in a behavioural manor. See [Behavioural Driven Development](../../docs/Test.md#style-of-test).
 
 ## Expected files
 
--- a/client/Bootstrap/README.md
+++ b/client/Bootstrap/README.md
@@ -6,7 +6,7 @@ While it's own directory, the code in this directory should be tested and develo
 
 ## Test approach
 
-To be defined in https://github.com/strimzi/strimzi-ui/issues/7
+Bootstrap should be tested in a behavioural manor. See [Behavioural test style](../../docs/Test.md#behavioural-test-style).
 
 ## Expected files
 
--- a/docs/Test.md
+++ b/docs/Test.md
@@ -56,6 +56,13 @@ with React component code:
 - `client/Groups`
 - `client/Panels`
 
+### Server Testing
+
+The strimzi-ui server should be tested from outside the server - like how a
+user would use it. It should be behaviourally driven using _Jest_ and
+_Cucumber_ to talk to each enpoint and check that the response received is
+correct. The test code for the server is in the `server/test` directory.
+
 ### End to End testing
 
 End to End tests should exercise a user scenario - focus on the behaviour a
@@ -102,7 +109,8 @@ confirm the test is clean and will be reliable:
 In this repo, we aim to have 100% code coverage across all components. This code
 coverage should be achieved through useful functional/behavioural tests and
 not just units to increase the coverage. We want all lines to be verified
-through meaninful tests.
+through meaninful tests. Code coverage should be used as a tool to help spot
+where certain tests may have been overlooked.
 
 ## Accessibility testing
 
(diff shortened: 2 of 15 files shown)


docs/Test.md
```diff
@@ -0,0 +1,150 @@
...
+confirm the test is clean and will be reliable:
+
+- Date and time should be mocked/set to a known value
+- Tests should cause no side effects - they should create and delete any
+  resources they need in the scope of that test/test suite
+- Tests (regardless of style) should be structured in the following way:
+  1. Perform any required setup
+  2. Perform an action (or actions)
+  3. Assert the result
+  4. Perform any required tidyup
```
dlabaj: can we change tidyup to clean up

docs/Test.md
```diff
@@ -0,0 +1,150 @@
...
+- Validate we do not regress functionality or behaviour we care about
+
+We want to enable and achieve all of these goals in the UI testing done in this
+repo. To this end, we require, depending on the functionality being added or
+changed, different types of tests. The types of test expected, and the
+rationale for these are detailed below.
+
+## Style of test
+
+We have two styles of test - `Behavioural` and `Functional`.
```
dlabaj: Instead Behavioral style and Functional style do we just want to go with Behavior Drive Development and Test Driven Development?

docs/Test.md
```diff
@@ -0,0 +1,150 @@
...
+### Functional test style
+
+Code in the following directories are expected to be tested by `functional`
+test methods, using _Jest_ as a test runner, using standard Jest test assets,
+such as `describe`, `it` and `expect`:
+
+- `client/Hooks`
+- `client/Contexts`
+
+### Behavioural test style
```
dlabaj: Where are we putting the behavioral and functional tests?  Are they going to live alongside the components, elements, hooks, contexts, and panel code in src?   Or are we going to put them somewhere else?

dlabaj (commented):
A few comments.

docs/Test.md
JQrdan: Fair point, I'll update these to be more familiar

docs/Test.md
JQrdan: They'll go alongside each component as a regular test would. The split between behavioural/functional is more of a guidance of what approach should be taken for each component. I'll add a piece that says this is where they should go

docs/Test.md
```diff
@@ -0,0 +1,154 @@
...
+render = (children) =>
+  renderWithContextProviders(children, {}, [
+    { provider: MyProvicer1, value: providerValue1 },
+    { provider: MyProvider2, value: providerValue2 },
+  ]);
+```
+
+This can be defined in a test file so that tests can still use a `render`
+function in the same way they usually would with rtl but these renders will be
+wrapped in the providers defined in the above function.
```
matthew-chirgwin: Would suggest linking off to the `utils/test/README` here, and having links in that README to specific helpers (like the one here for contexts). I can see this list growing, making this document harder to follow.

client/Bootstrap/README.md
```diff
@@ -6,7 +6,7 @@ While it's own directory, the code in this directory should be tested and develo
 
 ## Test approach
 
-To be defined in https://github.com/strimzi/strimzi-ui/issues/7
+Bootstrap should be tested in a behavioural manor. See [Behavioural Driven Development](../../docs/Test.md#style-of-test).
```
matthew-chirgwin: I think it would be worth adding here that the tests for `Bootstrap` level components (and the whole UI more generally) are kept in the `e2e` folder (with a link to it).

e2e/README.md
```diff
@@ -0,0 +1,3 @@
+# End to End testing
+
+Please find documentation for E2E testing [here](../../docs/Test.md#end-to-end-testing).
```
matthew-chirgwin: Would suggest having a brief summary here of what this directory contains (along with links to sub folder readmes)

docs/Test.md
```diff
@@ -0,0 +1,154 @@
...
+  resources they need in the scope of that test/test suite
+- Tests (regardless of style) should be structured in the following way:
+  1. Perform any required setup
+  2. Perform an action (or actions)
+  3. Assert the result
+  4. Perform any required clean up
+
+## Code coverage
+
+In this repo, we aim to have 100% code coverage across all components. This code
```
matthew-chirgwin, line 110: I had an idea around coverage: based on the files changed, run just related tests via Jest (for example), and get coverage on just those results. Ie, tests related to the change made to check coverage - no chance of having another test happen to 'run over' lines.

I can see this being non trivial, but will mean an area can be fully excercised by it's tests alone.

utils/test/README.md
```diff
@@ -0,0 +1,3 @@
+# Test Utils
+
+This directory contains all files that provide utility to all tests. See [Test utilities](../../docs/Test.md#test-utilities) .
```
matthew-chirgwin: See https://github.com/matthew-chirgwin/strimzi-ui/tree/master/utils/build for example of how this file could look

matthew-chirgwin (changes requested):
Looks good! Other than comments already left, just a few things which would suggest we add:

- Add a link to `docs/README.md` to the Test doc
- Add a link on the main repo README to the test document

client/Bootstrap/README.md
JQrdan: Happy to say bootstrap needs to only be tested through E2E as that is much more meaningful than trying to make fiddly unit tests. Hesitant about the navigation pieces being E2E only but willing to at least try, depending on its implementation. @dlabaj do you have any thoughts on this?

docs/Test.md
JQrdan: For now, I'll just expand on what we believe the use of code coverage is. Further investigation on selective test coverage is needed before this can be fleshed out. I have raised an investigation point on the project board https://github.com/orgs/strimzi/projects/2#card-46309994.

docs/Test.md
```diff
@@ -0,0 +1,143 @@
...
+assets, such as `describe`, `it` and `expect`, and _RTL_ to mount and interact
+with React component code:
+
+- `client/Elements`
+- `client/Groups`
+- `client/Panels`
+
+### Server Testing
+
+The strimzi-ui server should be tested from outside the server - like how a
```
matthew-chirgwin: `The strimzi-ui server should be tested from outside the server - like how a user would use it`. I think this might need a slight grammar tweak :)  Honestly, I am not sure if this sentence adds much. The main thing to know is the server code is/should be tested in a behavioural manner, as a user would acessing it (indirectly) via the UI. I think the second sentence (with a few extra words) could do the job.

matthew-chirgwin (commented):
LGTM, bar a question around grammar - thanks :)

matthew-chirgwin (approved):

scholzj (approved):

matthew-chirgwin:
Given there have been no further comments, merging this now.

PR #32, opened 2020-10-02
feat: add dependency approach doc

### LINKED ISSUE #31: Architecture: dependency approach
This issue is to add to the Architecture document a section around npm dependency mangement, and that we could (and I think should) place dependencies in either `dependencies` or `devDependencies` with care. 

In my opinion, any dependency which is used directly, or is responsible for generating/producing code which is built and used at runtime should be present `dependencies` . This means in the case where a security issue is found, we can quickly identify if it effects the 'shipped'/built code. This can be then leverdged via tools like `npm audit` to provide further details on just the `production` dependencies for instance.

matthew-chirgwin:
- adds section to cover dependency approach

Closes: #31

Signed-off-by: Matthew Chirgwin <chirmatt@uk.ibm.com>


docs/Architecture.md
```diff
@@ -278,6 +279,23 @@ For this capability to work, [file name conventions](#file-name-conventions) nee
 
 _Note_: Implementation of this will follow in a future PR.
 
+#### Dependency management
+
+This UI will make use of a number of third party dependencies to operate at develop, test, and runtime. The will range from helper tools such as Webpack, to frameworks, such as React. The `package.json` file/schema defines three types of dependency, `dependencies`, `peerDependencies` and `devDependencies`. Depending on which section a dependency sits, it will be treated differently at build time, and moreover, imply/have a different contextual meaning:
```
JQrdan: The second sentence reads weird to me

docs/Architecture.md
```diff
@@ -278,6 +279,23 @@ For this capability to work, [file name conventions](#file-name-conventions) nee
...
+
+The Strimzi-ui will not be published. Instead, it will consume a number of packages, and be built to provide a UI. Some of these packages will be used and shipped directly, but some will also generate and polyfill code in this repository to produce the the built UI. Thus, in the event of a security issue, we need to understand if the issue relates in any way to what we ship. We therefore categorise our dependencies as follows:
+
+- `dependencies` are any dependency which are either shipped bundled in the built output via direct usage (eg React), included in the dockerfile which will run the UI (eg Express) or generate output that is then built and shipped (ie babel transforms/plugins, Webpack plugins etc)
+- `devDependencies` are used to support the development and test of the UI, but that are not shipped. Eg Webpack, Jest, eslint
+
```
JQrdan: E.g. and I.e. instead of eg and ie

JQrdan (changes requested):
Sounds good to me, just some minor changes needed

JQrdan (approved):
LGTM

scholzj (approved):

docs/Architecture.md
```diff
@@ -167,9 +168,9 @@ Further details will be added to this section once https://github.com/strimzi/pr
...
+A UI typically allows a user to interact with resources. These resources are types of known entity (E.g. a Topic), which typically can be created, read, updated or deleted. Similarly, entities are often protected by authorization controls, restricting a user's access or ability to create, read, update or delete a given named entity or entity type. In the case of authorization for example, different mechanisms may be used to grant or reject the ability to create for instance. Equally, the ability to 'retrieve' an entity may vary (i.e retrieve the whole entity, or just a particular attribute) by use case. From a UI point of view, the 'how' of an operation on an entity is not important. What is important is if we can or cannot, and how the UI should ultimately respond to that.
```
dlabaj: I think we can get rid of the word Similarly and just use "Entities are often..."

dlabaj: Same with Equally... could just say "The ability..."

docs/Architecture.md
```diff
@@ -278,6 +279,23 @@ For this capability to work, [file name conventions](#file-name-conventions) nee
 
 _Note_: Implementation of this will follow in a future PR.
 
+#### Dependency management
+
+This UI will make use of a number of third party dependencies to operate at develop, test, and runtime. This will range from helper tools such as Webpack, to frameworks, like React. The `package.json` file/schema defines three types of dependency, `dependencies`, `peerDependencies` and `devDependencies`. Depending on which section a dependency sits, it will be treated differently at `npm publish` (build) time:
```
dlabaj: We could get rid of the `npm publish` part since we are not publishing this package, and change it to `npm build` everywhere instead.

docs/Architecture.md
```diff
@@ -278,6 +279,23 @@ For this capability to work, [file name conventions](#file-name-conventions) nee
...
+- `devDependencies` are used to support the development and test of a package, but are not shipped
+
+Tooling, such as [`npm audit`](https://docs.npmjs.com/cli/audit) make use of/utilise these sections to report when a dependency has a known vulnerability, and can thus call out if an issue is shipped by this package, assuming that dependencies are correctly catagorised based on their usage. Given the nature of JavaScript and where it is used, having these tools report issues when they occur is incredibly valuable.
+
+The Strimzi-ui will not be published. Instead, it will consume a number of packages, and be built to provide a UI. Some of these packages will be used and shipped directly, but some will also generate and polyfill code in this repository to produce the the built UI. Thus, in the event of a security issue, we need to understand if the issue relates in any way to what we ship. We therefore categorise our dependencies as follows:
```
dlabaj, line 393: Can we consolidate this section with the one above, so we only have 3 bullets instead of 5 bullets (dependencies, devDependencies, and peerDependencies)?

docs/Architecture.md
```diff
@@ -278,6 +279,23 @@ For this capability to work, [file name conventions](#file-name-conventions) nee
...
+
+- `dependencies` are any dependency which are either shipped bundled in the built output via direct usage (E.g. React), included in the dockerfile which will run the UI (E.g. Express) or generate output that is then built and shipped (I.e. babel transforms/plugins, Webpack plugins etc)
```
dlabaj, line 395: I noticed `npm audit` does audit devDependencies as well.  Wondering if we can get rid of the generate output part of this.  I believe having them in devDependencies will work correctly for us.

dlabaj (commented):
Thanks for submitting this.  A few comments / questions.

docs/Architecture.md
matthew-chirgwin: Both updated - good suggestions :)

docs/Architecture.md
matthew-chirgwin: This was intended to be a quick summary of the normal npm approach to dependencies, with the following sentences/paragraphs explaining the difference in approach we take. I will update the wording to be more clear here.

docs/Architecture.md
matthew-chirgwin: `npm audit` is one tool which can list both normal and dev dependencies. There are others which cannot/do not however (just reading the dependencies section). Plus, some generated code is included in/amongst the code we write in the built package, as if we imported it.

matthew-chirgwin:
Thank you for the comments @dlabaj / @JQrdan - I have rebased and updated the PR given these. If there are no more comments I will look to merge this tomorrow

docs/Architecture.md
dlabaj: okay then this is fine.  Thanks for the clarification.

docs/Architecture.md
dlabaj: No worries.  I just didn't want someone to get confused reading this.  Thanks again.

PR #29, opened 2020-10-01
Typescript React Recommendations

dlabaj:
Closes issue #19

Signed-off-by: Donald Labaj <dlabaj@redhat.com>

PR #46, opened 2020-10-21
feat: i18n support

### LINKED ISSUE #21: Translation tooling/approach
The UI should be fully translatable from day 0. This issue is to look into either writing our own translation tooling or using a 3rd party library, and writing supporting documentation to how it works/should be used across the Strimzi-ui

nictownsend:
- react-i18next for translation
- example component
- example test
- react-i18n mock and helpers

Contributes to: strimzi/strimzi-ui#21


__mocks__/react-i18next.tsx
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+/* eslint-disable react/prop-types */
```
JQrdan: Given that typescript covers our prop-types needs, should we just turn off this linting rule?

build/webpack.common.js
```diff
@@ -128,38 +131,54 @@ const withImageModuleLoader = returnModuleRuleWithConfig(
...
-    alias: webpackAliases,
+const withTypescriptModuleLoader = returnModuleRuleWithConfig(
+  {
+    test: /\.tsx?$/,
+    exclude: /node_modules/,
   },
-  ...customConfigurationForBuildMode,
-});
+  ['ts-loader']
+);
```
JQrdan: Do `.ts` files need to be passed through this as well? Or is it something specific to`.tsx`?

client/Bootstrap/index.tsx
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import '@/Styling';
```
JQrdan: I know you've mentioned the `@/` style before, is there a reason for the `/`? I've seen some examples where it's just `@module` for local aliases which blends in nicely with some `node_modules` imports

client/Pages/Home/Home.spec.tsx
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+/* eslint-disable react/jsx-key */
+jest.mock('react-i18next');
+import { render } from '@testing-library/react';
+import React from 'react';
+import { Home } from '.';
+import { translate, translateWithFormatting } from '@/utils/test/i18n';
```
JQrdan: Could the utils modules become (for example) `test-utils` for these ones and then appropriate names for others? Or does these name get generated by the file structure?

client/Pages/Home/Home.spec.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+it('should render', () => {
```
JQrdan: Can you wrap this in a describe so there's a bit more context when these fail

client/Pages/Home/Home.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import image from '@/Images/logo.png';
+import React, { ChangeEvent, FunctionComponent } from 'react';
+import { useTranslation, Trans } from 'react-i18next';
+import i18next from 'i18next';
+interface Button {
+  prop: string;
+}
+
+// eslint-disable-next-line react/prop-types
```
JQrdan: Marker for prop-types comment if we do remove it

client/Pages/Home/Home.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const changeLang = (e: ChangeEvent<HTMLSelectElement>) => {
+  document.querySelector('html')?.setAttribute('lang', e.target.value);
+  i18next.changeLanguage('');
+};
+
+const langs = ['en', 'de'];
+
+const Home: FunctionComponent = () => {
+  const { t } = useTranslation();
```
JQrdan: Single letter variable!?!?

__mocks__/react-i18next.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { translate, translateWithFormatting } from '@/utils/test/i18n';
+import React, { FunctionComponent } from 'react';
+
+const Trans: FunctionComponent<{ i18nKey: string }> = ({
```
JQrdan, line 22: Can this component be called `Translate` instead

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
+This project uses `i18next` with `react-i18next` to provide internationalization of content. Human readable content should not be used directly in React components. Instead, they should be looked up using `react-i18next` - which will ensure the content is localized. The library also supports interpolation of dynamic values and formatting of the output. `eslint-plugin-i18next` is used to ensure that JSX cannot be written that contains text without a translation.
+
+## Usage
+
+The `useTranslation` hook provides a `t(key, options)` function - that takes a translation key and (optional) options and returns the translated content. See [https://www.i18next.com/translation-function/essentials](https://www.i18next.com/translation-function/essentials) for more details. An object hash containing values for interpolation can be supplied as the second parameter to `t` instaed of a full options object.
```
JQrdan:
```suggestion
The `useTranslation` hook provides a `t(key, options)` function - that takes a translation key and (optional) options and returns the translated content. See [https://www.i18next.com/translation-function/essentials](https://www.i18next.com/translation-function/essentials) for more details. An object hash containing values for interpolation can be supplied as the second parameter to `t` instead of a full options object.
```

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+```
+import {useTranslation} from 'react-i18next';
+...
+export const myComponent = () => {
+    const {t} = useTranslation();
+    return (
+        <>
```
JQrdan, line 17: Empty brackets?

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
+This project uses `i18next` with `react-i18next` to provide internationalization of content. Human readable content should not be used directly in React components. Instead, they should be looked up using `react-i18next` - which will ensure the content is localized. The library also supports interpolation of dynamic values and formatting of the output. `eslint-plugin-i18next` is used to ensure that JSX cannot be written that contains text without a translation.
+
+## Usage
+
+The `useTranslation` hook provides a `t(key, options)` function - that takes a translation key and (optional) options and returns the translated content. See [https://www.i18next.com/translation-function/essentials](https://www.i18next.com/translation-function/essentials) for more details. An object hash containing values for interpolation can be supplied as the second parameter to `t` instaed of a full options object.
+
+```
```
JQrdan: We should probably start putting the language type when we do code blocks so that the syntax highlighting is correct

client/i18n/README.md
JQrdan: I see this is just used to show that there's something wrapping the content, I'm not too fussed about this

e2e/plugins/index.ts
```diff
@@ -4,7 +4,7 @@
  */
 import cucumber from 'cypress-cucumber-preprocessor';
 
-module.exports = (on: Function) => {
+module.exports = (on: (a: unknown, b: unknown) => unknown) => {
```
JQrdan: a and b?

linting/eslint.config.js
```diff
@@ -23,7 +13,7 @@ module.exports = {
     browser: true,
     node: true,
     es6: true,
-    'jest/globals': true,
+    jest: true,
```
JQrdan: I'm guessing the `/globals` part isn't needed for eslint to be happy with jest code?

package.json
```diff
@@ -1,48 +1,45 @@
 {
```
JQrdan, line 1: What's the reason for the re-jig of this file?

tsconfig.json
```diff
@@ -1,9 +1,19 @@
...
-    "target": "es5",
-    "lib": ["es5", "dom"],
-    "types": ["node", "jest"]
+    "esModuleInterop": true,
+    // "target": "esnext",
+    "lib": ["esnext", "dom"],
+    "types": ["node", "jest", "@testing-library/jest-dom"],
+    "allowJs": true,
+    "resolveJsonModule": true,
+    // "checkJs": true,
```
JQrdan: ?

utils/test/i18n/i18n.util.ts
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { ReactNode, ReactElement } from 'react';
+import { isElement } from 'react-dom/test-utils';
+
+const stringify: (node: ReactNode) => string = (node) => {
+  if (node === undefined || node === null) {
```
JQrdan: Can this just be replaced with `if (!node) {...}`

utils/test/i18n/i18n.util.ts
```diff
@@ -0,0 +1,70 @@
...
+    return '';
+  }
+
+  if ((node as ReactElement).props) {
```
JQrdan, line 12: I'm interested to know why this needs to be casted?

JQrdan: If it definitely is needed, could a casted variable be saved and re-used instead of casting each time

utils/test/i18n/i18n.util.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  return `key:"${key}"${inserts ? ` inserts:${inserts}` : ''}`;
+};
+
+export const translateWithFormatting: (
+  key: string,
+  children:
+    | Record<string, string>
+    | ReactNode
+    | [Record<string, string> | ReactNode]
```
JQrdan: This seems iffy to me, that's a lot of types

JQrdan (changes requested):
Overall looks good! Although I'm confused as to why the `<Trans>` component is needed alongside a hook instead of just using a hook?

__mocks__/react-i18next.tsx
nictownsend: Good point, will add

build/webpack.common.js
nictownsend: I believe `.tsx?` regex catches both `.ts` and `.tsx`

client/Bootstrap/index.tsx
nictownsend: So I was thinking of it more as a convention for "local" modules - as lots of node modules use `@module` for namespacing. Happy to remove the `/` though if we want to look more like all other modules.

client/Pages/Home/Home.spec.tsx
nictownsend: at the moment - module mapping is essentially replacing `../../../..` to get to root with `@/` and as such for now the aliases aren't manual but instead based off file path.

client/Pages/Home/Home.spec.tsx
nictownsend: Fair point, will do.

__mocks__/react-i18next.tsx
nictownsend: I can do for the local const, but the export will need to be called `Trans` as that's the `react-i18next` api

client/i18n/README.md
nictownsend: It's shorthand for `<React.Fragment>` too

client/i18n/README.md
nictownsend: Will do

e2e/plugins/index.ts
nictownsend: Linting complained that `Function` was not acceptable, so I was giving default var names for the type definition (but will add more sensible names)

build/webpack.common.js
JQrdan: Apologies, I missed the `?`

tsconfig.json
nictownsend: Commented out for discussion - I was finding that compiler was not happy with the options set (including validating JS files as TS). But open for discussion.

client/i18n/README.md
JQrdan: TIL

utils/test/i18n/i18n.util.ts
nictownsend: Fair point

utils/test/i18n/i18n.util.ts
nictownsend: From my research, this is the way to do `instanceof` checks in TS - plus I think it would be more memory efficient than creating casted variables?

utils/test/i18n/i18n.util.ts
nictownsend: I'll play around and see if `Record<string,string>` can be satisfied by `ReactNode`

nictownsend:
> Overall looks good! Although I'm confused as to why the `<Trans>` component is needed alongside a hook instead of just using a hook?

So I can't speak to the `react-i18n` library's design - but the hook will only ever return strings, whereas `<Trans>` will render the contents (so for e.g a translations of `this text is <strong>bold</strong>` will be the string with the hook, but html with `<Trans>`.

package.json
nictownsend: Alphabetisation :)

__mocks__/react-i18next.tsx
```diff
@@ -0,0 +1,20 @@
...
+  i18nKey,
+  children,
+}) => <div>{translateWithFormatting(i18nKey, children)}</div>;
+
+const useTranslation = () => {
+  return { t: translate };
```
matthew-chirgwin: any reason why we could not export `translate` as translate? I think the name should at least reflect what `t` is/is for

build/webpack.common.js
```diff
@@ -128,38 +131,54 @@ const withImageModuleLoader = returnModuleRuleWithConfig(
...
+
+// exported helper function - returns common configuration, merged via object spread. Provided options take precedence
+const returnBasicConfigMergedWith = (customConfigurationForBuildMode = {}) =>
+  merge(
+    {
+      entry: `${BOOTSTRAP_DIR}/index.tsx`,
```
matthew-chirgwin, line 162: It isn't merged yet, but when the server builds, it has `.ts` entry point - should/would this apply to client webpack build configs only? (Happy to change in my PR if this makes it in first)

client/Bootstrap/index.tsx
matthew-chirgwin: I dont mind the `/` or `@/` myself, as long as the `@/` is documented as being 'local' code

client/Bootstrap/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Home } from '@/Pages';
+import './i18n';
+import ReactDOM from 'react-dom';
+import React from 'react';
+
+const onLoad: () => void = () => {
```
matthew-chirgwin: Still on browser load? Was expecting index would be a script which fires as soon as it can to (start) rendering the UI

client/Pages/Home/Home.spec.tsx
```diff
@@ -0,0 +1,44 @@
+/*
```
matthew-chirgwin: `Pages` is meant for Page metadata - not components. Can this be moved to `Panels` (I know its only placeholder for now). 

Same is true for other files relating to `Home`

client/Pages/Home/Home.tsx
matthew-chirgwin: Yeah - I really dont like it haha :)

client/custom.d.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+declare module '*.png';
```
matthew-chirgwin: Any reason why this isnt in the images folder (to keep with the 'code' it represents)?

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+### Providing translation values
+
+`Bootstrap/i18n.ts` configures the `i18next` library to load translations from `./index.ts`.
+
+`index.ts` exports an Object containing translations from `<language>.ts`.
+
+`<language>.ts` contains the following shape - where `{{key}}` is substituted using a matching key in the inserts passed to `t()`:
```
matthew-chirgwin: Would a type/interface be possible to help enforce/guide usage?

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
+This project uses `i18next` with `react-i18next` to provide internationalization of content. Human readable content should not be used directly in React components. Instead, they should be looked up using `react-i18next` - which will ensure the content is localized. The library also supports interpolation of dynamic values and formatting of the output. `eslint-plugin-i18next` is used to ensure that JSX cannot be written that contains text without a translation.
```
matthew-chirgwin, line 3: Uber picky - worth adding a title for nicer formatting (`# internationalisation (i18n)`)?

client/i18n/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Usage
+
+The `useTranslation` hook provides a `t(key, options)` function - that takes a translation key and (optional) options and returns the translated content. See [https://www.i18next.com/translation-function/essentials](https://www.i18next.com/translation-function/essentials) for more details. An object hash containing values for interpolation can be supplied as the second parameter to `t` instead of a full options object.
```
matthew-chirgwin, line 7: What happens if I use/reference a translation which isnt available in all language files? I also assume all translations must exist in the base (en) file?

e2e/plugins/index.ts
matthew-chirgwin: I ran into this too and went with `(on: (evt: string, callback: () => void) => void)`. Not a must change, but interesting to see what you came up with!

package.json
matthew-chirgwin: No objection - but boy will this be a fun rebase to sort!

package.json
```diff
@@ -1,48 +1,45 @@
...
-  "license": "Apache-2.0",
+  "description": "Strimzi UI provides a way for managing Strimzi and Kafka clusters (+ other components) deployed by it using a graphical user interface.",
+  "sideEffects": false,
+  "version": "0.0.1",
   "bugs": {},
-  "homepage": "https://github.com/strimzi/strimzi-ui#readme",
   "dependencies": {
     "@babel/core": "^7.10.5",
     "@babel/preset-env": "^7.10.4",
+    "@typescript-eslint/eslint-plugin": "^4.4.1",
```
matthew-chirgwin, line 46: is this rule checking or formatting?

package.json
```diff
@@ -1,48 +1,45 @@
...
     "babel-loader": "^8.1.0",
     "compression-webpack-plugin": "^4.0.0",
+    "eslint": "^7.6.0",
```
matthew-chirgwin: as a `dependencies`?

package.json
```diff
@@ -1,48 +1,45 @@
...
+    "file-loader": "^6.0.0",
     "html-webpack-plugin": "^4.3.0",
+    "http-proxy": "^1.18.1",
+    "i18next": "^19.8.2",
+    "i18next-browser-languagedetector": "^6.0.1",
+    "license-check-and-add": "^3.0.4",
+    "lint-staged": "^10.2.11",
```
matthew-chirgwin, line 58: as a `dependencies`?

package.json
```diff
@@ -1,48 +1,45 @@
...
+    "lodash.merge": "^4.6.2",
     "mini-css-extract-plugin": "^0.9.0",
+    "npm-run-all": "^4.1.5",
```
matthew-chirgwin: as a `dependencies`?

package.json
```diff
@@ -1,48 +1,45 @@
...
     "optimize-css-assets-webpack-plugin": "^5.0.3",
+    "prettier": "^2.0.5",
     "react": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "react-i18next": "^11.7.3",
     "sass": "^1.26.10",
+    "sass-loader": "^9.0.2",
+    "style-loader": "^1.2.1",
+    "stylelint": "^13.6.1",
```
matthew-chirgwin, line 69: as a `dependencies`?

utils/test/i18n/i18n.util.ts
matthew-chirgwin: `node` is typed as a `ReactNode` (on 7), then used as a `ReactElement` here. The `as` is a cast, but as a hint to TS. You could do `const foo = {} as ReactElement` and from a TS POV call/use foo as if it were a `ReactElement` (when it is not). This is needed when TS cannot (despite best effort) figure out what X is.

Do you know what the relationship is between `ReactElement` and `ReactNode` @nictownsend ? I assume `ReactElement` is a `ReactNode`, but we props are only typed on the ReactElement?

matthew-chirgwin (changes requested):
Looks good! Have a couple of questions and comments - and can we please not use `t`? :)

e2e/plugins/index.ts
nictownsend: https://github.com/cypress-io/cypress/blob/5c01832883e92de7a596e5d80a844cc04cf3ae32/cli/types/cypress.d.ts#L5002

__mocks__/react-i18next.tsx
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { translate, translateWithFormatting } from '@/utils/test/i18n';
```
nictownsend: @JQrdan @matthew-chirgwin - https://react.i18next.com/latest/usetranslation-hook - we can either use named export `t` or we can use the `[t, i18next]` version and rename it on every import.

From mock perspective, I'll update it to return both APIs for safety. But I'll have to export `t` as function regardless.

build/webpack.common.js
nictownsend: I'm guessing we'll need separate config for server (`target` in particular)

client/Bootstrap/index.tsx
nictownsend: I just slotted in with the existing page. Happy to remove if that was just there for initial purposes.

client/Pages/Home/Home.spec.tsx
nictownsend: Yep, will move.

client/Pages/Home/Home.spec.tsx
nictownsend: I'll start a thread on strimzi-ui regarding aliasing.

client/Pages/Home/Home.tsx
nictownsend: As mentioned in the mock - we can either use the `[]` return value or destructure the named variable.

But, the `<Trans>` component does have a prop called `t` - so we can't avoid it entirely.

I'm happy to destructure, but it does mean in future a find/replace would be harder to achieve.

client/custom.d.ts
nictownsend: Fair point, will move.

client/i18n/README.md
nictownsend: Yep will do!

client/i18n/README.md
nictownsend: If you're using `t` function, it falls back to base language, or the key if base is missing too.

If you're using `trans` component, it renders the children if it cannot fall back.

linting/eslint.config.js
nictownsend: Reverted, not sure why I changed it

package.json
nictownsend: Sorry all! :)

package.json
nictownsend: We run `--fix` so my assumption was it would manipulate code

package.json
nictownsend: All dev dependencies I moved if their usage manipulated the output

nictownsend: I think I'd prefer if *all* compile time were dev dependencies - but as we've previously discussed, anything changing the output would need to be flagged for security scans (and moving it to dependencies allows quick flagging)

utils/test/i18n/i18n.util.ts
nictownsend: Exactly that. Function components `children` are `ReactNode` - which can be string, number, boolean, ReactNodeArray, ReactElement. So to get props, you have to be working against a ReactElement type.

I'm using the `as` rather than the cast to avoid new object creation but if we think it cleaner I'm happy to do the reassign cast.

client/Bootstrap/index.tsx
matthew-chirgwin: As we are starting to add actual react code (even placeholder code at that), I think it is time to switch off browser load - it should fire as soon as it is loaded.

build/webpack.common.js
matthew-chirgwin: Absolutley - I wonder as the server and client will be different here is the merit in removing entry from common (as it now wont be)?

client/i18n/README.md
matthew-chirgwin: Worth mentioning that here?

docs/Build.md
```diff
@@ -137,6 +82,36 @@ These options will all be provided to the TerserPlugin via it's constructor.
 | terserOptions.keep_fnames     | `true`                                  |                   Keep original function names |
 | terserOptions.mangle.safari10 | `true`                                  |               Works around known Safari issues |
 
+##### Swappable view layers
+
+**TODO - this approach needs to change with use of NormalModuleReplacementPlugin**
```
matthew-chirgwin, line 89: Would you mind raising an issue for this? Out of scope for this PR, but should be done sooner rather than later

package.json
matthew-chirgwin: eslint is the runtime, which rule checks. Plugins modify. Worth confirming eslint itself doesnt bundle plugins. If it does, then it should be a dep by our definition

tsconfig.json
```diff
@@ -1,9 +1,19 @@
...
+    "esModuleInterop": true,
+    "target": "ES2019", //Until webpack can handle JS optional chaining syntax
+    "lib": ["esnext", "dom"],
+    "types": ["node", "jest", "@testing-library/jest-dom"],
+    "allowJs": true,
+    "resolveJsonModule": true,
+    "moduleResolution": "node",
+    "baseUrl": "./",
+    "paths": {
+      "@/Styling": ["client/Bootstrap/Styling.scss"],
```
matthew-chirgwin: As per the Webpack approach, Styling was a relative alias - ie it resolved to a `Styling.scss` in that directory - not a top level file. I still have this as a relative alias from webpack in my server work, and have no path entry for it in tsconfig

utils/test/i18n/i18n.util.ts
matthew-chirgwin: Tell me if it wont work, but why not treat `node` as a ReactElement from line 7 (ie change the typing of the function)?

client/Bootstrap/index.tsx
dlabaj: If we are going to use @ for local I would prefer it be more specific such as _@strimzi-ui/Styling_

dlabaj: I don't think we need @/Styling it's only imported once during Bootstrapping.  It makes more sense to me to do `import './Styling';` in this case.

tsconfig.json
dlabaj: I would prefer we get ride of declaring this path for a single styling file.  Can we also make Styling.scss lower case style.scss.

tsconfig.json
```diff
@@ -1,9 +1,19 @@
...
+      "@/*": ["*", "client/*", "server/*"] //https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping
```
dlabaj: I think this should be broken down more.  There should be on for the `@strimzi-ui/* ` and one for the ui server `@strimzi-ui-server/*`. I would think we would want to restrict access to ui code in the server, and server code in the ui.  If we need common code to share between them we should make a common top level folder.

docs/Build.md
nictownsend: I believe it's in @dlabaj PR for typescript?

utils/test/i18n/i18n.util.ts
nictownsend: `ReactNode` contains `string|number|boolean|{}` and a fair few more types. Ideally I don't want to be type checking against all of those - so instead, pass in `ReactNode`, check if it has props as `ReactElement` and stringify that - otherwise, `toString()`.

It's missing a few of the more complex types from `ReactNode` like fragments, node arrays, portals. But for the sake of the translation, this seemed like the best way (as I don't expect anyone to pass anything other than primitives or a React Element to the `<Trans>` component.

tsconfig.json
nictownsend: It'll need to be in tsconfig for jest purposes

client/Bootstrap/i18n.ts
```diff
@@ -0,0 +1,33 @@
+/*
```
dlabaj: Can we move this file into i18n folder.

client/Panels/Home/Home.tsx
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import image from '@/Images/logo.png';
+import React, { ChangeEvent, FunctionComponent } from 'react';
+import { useTranslation, Trans } from 'react-i18next';
+import i18next from 'i18next';
+interface Button {
```
dlabaj: Would call this ButtonProps and make it a type.

docs/Build.md
dlabaj: yes it's in typescript pr

dlabaj (changes requested):
A few changes other than that looks good.  Thanks.

tsconfig.json
nictownsend: Removed with discovery that TS performs module lookups relative to `baseUrl` as well as `node_modules`

.github/actions/bundle/index.js
```diff
@@ -47,9 +47,10 @@ const createTableHeader = (tableHeadings) => {
 };
 
 const createBundleReportTable = (bundleReport, bundleSizes) => bundleReport.reduce((previousbundleText, bundle) => {
-  bundleSizes[bundle.label].current = bundle.parsedSize;
-  const currentSizeBytes = bundleSizes[bundle.label].current;
-  const masterSizeBytes = bundleSizes[bundle.label].master;
+  const bundleSize = bundleSizes[bundle.label] || {master: 0};
```
nictownsend, line 50: For any cases where a new bundle has been created (webpack)

build/webpack.common.js
```diff
@@ -39,7 +38,7 @@ const withHTMLPlugin = returnPluginWithConfig(HtmlPlugin, {
   template: `${BOOTSTRAP_DIR}/index.html`, // source template
   title: UI_TITLE, // HTML title
   favicon: `${IMAGES_DIR}/favicon.ico`, // favicon for this page
-  inject: 'head', // any/all built content linked into HTML head element
+  inject: true,
```
nictownsend, line 41: If script loads from `<head>` it's before the DOM exists (so react can't mount elements)

test_common/jest.config.js
```diff
@@ -37,27 +15,15 @@ const config = {
...
-  moduleNameMapper: {
-    ...jestModuleMapper,
-  },
-  setupFiles: ['<rootDir>/test_common/jest_cucumber_support/index.ts'],
-  globals: {
-    'ts-jest': {
-      tsconfig: '<rootDir>/test_common/tsconfig.test.json',
-    },
-  },
+  coverageProvider: 'v8',
```
nictownsend, line 18: So...

If you use the default `babel` provider - it will not know how to provide coverage to the two projects as they have different tsconfig (one needs to support jsx + esnext, other is older js module).

If you set `v8` - it is able to hook into the compiled typescript itself. Coverage numbers seem to possibly be based upon the transpiled JS - but, I've found that if it says 100% line coverage it's telling the truth, but I don't know if the branches section is misleading.

docs/Build.md
matthew-chirgwin: Question for you both @dlabaj / @nictownsend - I have just been looking at #29 again. I can see the section that talks about file name converntion, but cannot see any mention nor integration of `NormalModuleReplacementPlugin` to enable the swapping of view layer used. What is the plan here? As a starter, I would suggest getting #29 in first, given its age/it is more doc/configuration based, where as this adds implementation of i18n as well.

matthew-chirgwin (changes requested):

build/webpack.common.js
matthew-chirgwin: TIL - can I catch up with you re this/the windows involved?

cypress.json
```diff
@@ -0,0 +1 @@
+{}
```
matthew-chirgwin: What is this for?

linting/eslint.config.js
```diff
@@ -35,14 +39,13 @@ module.exports = {
...
   // detect and use the version of react installed to guide rules used
   settings: {
     react: {
       version: 'detect',
     },
   },
   rules: customRules,
-  // ignore built output
-  ignorePatterns: ['**/dist'],
+  overrides: [tsConfig],
```
matthew-chirgwin, line 50: Much nicer 👍

package.json
```diff
@@ -23,40 +25,50 @@
     "test:e2e:open": "npm-run-all clean build --parallel --race start:e2eserver cy:open",
     "licence:check": "license-check-and-add --f ./linting/license-check-and-add.config.json check",
     "licence:add": "license-check-and-add --f ./linting/license-check-and-add.config.json add",
-    "lint": "npm-run-all 'lint:allfiles -- .' 'lint:js -- .' 'lint:ts -- .' 'lint:styling -- \"**/*.(s)css\"' 'lint:format -- .'",
+    "lint": "npm-run-all 'lint:allfiles -- .' 'lint:ts -- .' 'lint:styling -- \"**/*.(s)css\"' 'lint:format -- .'",
     "lint:allfiles": "npm-run-all 'licence:add -- {1}' --",
-    "lint:js": "eslint --config ./linting/eslint.js.config.js --fix --max-warnings 0 --ignore-path ./linting/.eslintignore",
-    "lint:ts": "eslint --config ./linting/eslint.ts.config.js --fix --max-warnings 0 --ignore-path ./linting/.eslintignore",
+    "lint:ts": "eslint --config ./linting/eslint.config.js --fix --max-warnings 0 --ignore-path ./linting/.eslintignore",
```
matthew-chirgwin: Being picky, as we now have a single config for all, can this be changed to not be file type specific - `lint:src` for instance? (With doc updates to match)

matthew-chirgwin (commented):
LGTM - just a few more questions/comments

client/i18n/README.md
```diff
@@ -0,0 +1,120 @@
...
+The `<Trans>` component also supports interpolation of JSX elements into the translated string to provide additional formatting (e.g class names, headers, links) around the string content. These elements must be passed as children to `<Trans>`. Dynamic content should be supplied after elements as singleton children as before.
+
+```tsx
+import {Trans, useTranslation} from 'react-i18next';
+export const myComponent = () => {
+    const {t: translate} = useTranslation();
+    return (
+        <>
+            <div><Trans t={translate} i18nKey="with.link><a href="."/></Trans></div>
+            <div><Trans t={translate} i18nKey="with.classname><div className="bold"/>{{topic: "mytopic"}}</Trans></div>
```
pmuir: Missing closing quotes here on `i18nKey` (and on preceeding and following line)

client/i18n/README.md
```diff
@@ -0,0 +1,120 @@
...
+            <div><Trans t={translate} i18nKey="with.custom><MyComp prop="value"/><SecondComp/></Trans></div>
```
pmuir: It would be nice to have a more complete example of passing in complex components like `MyComp` - but this can be a subsequent issue I think.

pmuir: Ah, actually this is covered in the next section.

client/i18n/README.md
```diff
@@ -0,0 +1,120 @@
...
+{
+  "with": {
+    "link": "This is a <0>link</0>",
+    "classname": "You have created topic <0>{{topic}}</0>",
+    "custom": "This is a custom component: <0/> with <1>another custom component</1>"
+  }
+}
+```
+
+### Notes/Findings with advanced formatting
```
pmuir, line 132: It would be good to write this in a more directional way to tell people what to do in these situations.

e2e/cypress-cucumber-preprocessor.config.js
```diff
@@ -3,5 +3,5 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 module.exports = {
-  stepDefinitions: 'e2e/features/step_definitions',
```
pmuir, line 6: I wasn't expecting to see the cucumber config change in a PR about i8ln - has something else crept in by mistake?

docs/Build.md
```diff
@@ -17,64 +17,9 @@ To maintain a small built bundle size, we take advantage of [Webpack's treeshaki
 
 Webpack (and it's plugins) are highly customisable. This section will detail the choices we have made around how and what is built, where it is built to, and the roles and responsibilities of the various plugins which enable this.
 
-#### Webpack aliases
```
pmuir, line 20: The changes to this file seem unrelated to the i8ln (and this PR is already massive).

pmuir (approved):
looks good, but for the sanity of reviewers, I think (in the future) we should try to split PRs more into logical areas - there is a lot in here that isn't related to i8ln.

nictownsend:
> looks good, but for the sanity of reviewers, I think (in the future) we should try to split PRs more into logical areas - there is a lot in here that isn't related to i8ln.

Agree @pmuir - I probably should have delivered the precursor `jest` work that was needed to then test the `i18n` work. I'm hoping this is the only time it's needed as framework will be there ready.

docs/Build.md
nictownsend: Agree - it was their prior to a refactor- will talk with @dlabaj to see what's ocurring

docs/Build.md
nictownsend: While I agree - they have come in because the changes needed to support `jest` with module mapping and typescript. So the "path/alias" resolution is now completed via `tsconfig` and such this section became out of date.

e2e/cypress-cucumber-preprocessor.config.js
nictownsend: Cypress tests were broken by the change to tsconfig to support react/jest. As part of getting the config working, I had changed to use cypress project to make the config cleaner (not needing to reference `e2e` in every path.

nictownsend: I agree - I may have taken boy scout principle (leave it cleaner than you started) slightly too far with this

nictownsend:
Closing due to community discussion - splitting into two PRs - one for adding jest tests for `client` code, one for adding i18n

PR #57, opened 2020-10-28
feat: add gql client

### LINKED ISSUE #55: Add GraphQL Client for transport layer
Implementation from details outlined in #38.

JQrdan:
Closes: #55


client/Queries/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This directory is home to all GraphQL queries. These should be imported by the appropriate models and exposed via a hook.
+
+## File Structure
+
+- Queries
+  - query set (e.g. topics)
+    - index.ts - containing all gql queries for that set
+
+## Example
```
matthew-chirgwin: Any thoughts on how/if these should be tested? I like the idea of keeping like queries together, but worry a query is changed for one use case, braking another. Is the expectation tests in say the topics page would fail in such a scenario?

client/Bootstrap/GraphQLClient/index.tsx
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { ApolloClient, InMemoryCache } from '@apollo/client';
+import { SubscriptionClient } from 'subscriptions-transport-ws';
+import { WebSocketLink } from '@apollo/link-ws';
+import config from '../../../config';
+
+const client = new SubscriptionClient(config.GRAPHQL_ENDPOINT, {
```
matthew-chirgwin: Is the intent to have one Client used across the whole UI? As mentioned, we will have two GQL sources to query, the Strimzi-api, as well as the config for the UI. Given these two endpoints cannot be federated (as they are pluggable), if we want to have a single client, we could use multiple links with context to direct queries to one or the other.

I would also suggest not referencing a config value for now. I would suggest this is hardcoded to `/api` (with a second link for config, again hardcoded to `/config`). Thoughts?

matthew-chirgwin (commented):
Looks good - have a couple of questions re approach

client/Bootstrap/GraphQLClient/index.tsx
pmuir: Given the BFF pattern, I think hardcoding the endpoints makes sense.

Multiple links makes sense, especially if we can use a directive to switch between them.

JQrdan: Agreed with multiple links, the query context can be used to decide where the request should be directed to

client/Bootstrap/GraphQLClient/index.ts
```diff
@@ -0,0 +1,28 @@
...
+const configLink = new HttpLink({ uri: '/config' });
+
+const splitLink = split(
+  (operation) => operation.getContext().purpose === 'config',
+  configLink,
+  subscriptionLink
+);
+
+const apolloClient = new ApolloClient({
+  link: splitLink,
```
pmuir: Can you add an example of the usage of the configLink to the `Queries/README.md`?

pmuir (approved):
Looks good

client/Queries/README.md
JQrdan: After some discussion, we've decided that queries shouldn't need testing themselves. The main argument being these could easily be .gql files. Re queries changing: this is a bridge to cross when we get to it although we all have some idea of what could happen when we get there

client/Bootstrap/GraphQLClient/index.ts
JQrdan: Added :)

nictownsend (approved):
LGTM

scholzj (approved):

Issue #44, opened 2020-10-19
Architecture (post proposal): Security


matthew-chirgwin:
To discuss/define:

- How authentication, authorization and sessions are handled in the UI
- CRUD model mentioned in https://github.com/strimzi/strimzi-ui/issues/12#issuecomment-679206635 and how this maps to authorization/roles

nictownsend:
Documentation delivered. Will raise additional issue (and an epic) for the follow on pieces.

PR #58, opened 2020-10-28
feat: introspection context

### LINKED ISSUE #39: Introspection Context
Issue to implement an Introspection Context, as described in https://github.com/strimzi/strimzi-ui/blob/master/docs/Architecture.md#introspection and https://github.com/strimzi/strimzi-ui/tree/master/client/Contexts/Introspect

- Requires GQL library ( #38 ), testing issues to be completed

pmuir:
Fixes #39


pmuir:
Adding this to get some early comments about whether this is the right direction or not.

This introspection handles the example in the readme. I need to add support for non Scalar types to the field checked. We possibly also want to consider passing some arguments to callbacks to allow them to evaluate whether the operation is supported. We may also want to check if particular types are attached to the query and mutation.

pmuir:
I had to import the graphql dependency but we can fix that up after we merge the GraphQL PR.

matthew-chirgwin (commented):
Thanks for this - looks exactly like what I was expecting. A few comments so far:

- Rather than implement .js files, could they be .ts (and define types/interfaces) for others to consume as appropriate?
- I assume you planning to add the React context producer/consumer pieces in a following commit?
  - As well as tests?
> We possibly also want to consider passing some arguments to callbacks to allow them to evaluate whether the operation is supported. 
- What scenarios do you see which may need this (in the scope of introspection)? I absolutley agree that the result of the introspection, security (#44) etc will need to be combined (as described in the page here: https://github.com/strimzi/strimzi-ui/tree/master/client/Pages - backendSupportFor in the case of introspection )

pmuir:
> Rather than implement .js files, could they be .ts (and define types/interfaces) for others to consume as appropriate?

Sure, I just used the js as existing files in `client` were js.

> I assume you planning to add the React context producer/consumer pieces in a following commit?

Right - I wanted to get feedback on actual meat of the code first.

I just pushed better field validation code.

> As well as tests?

Yes, I will write a simple harness that can read in sample introspection results and expectations, and process them. This will allow us to cover many different schemas/expectations. I'll start with the schema/expectations I've been validating with. 

I'll do this next.

> We possibly also want to consider passing some arguments to callbacks to allow them to evaluate whether the operation is supported.
> 
> What scenarios do you see which may need this (in the scope of introspection)? I absolutley agree that the result of the introspection, security (#44) etc will need to be combined (as described in the page here: https://github.com/strimzi/strimzi-ui/tree/master/client/Pages - backendSupportFor in the case of introspection )

I added some examples of this to the samples, but if we defined a `topicsByName` query, then we might want to check it's arguments to see if it supports certain arguments e.g.

```findByName: ({ queries }) => {
        return (
          queries['topicsByName'] !== undefined &&
          queries['topicsByName'].args.reduce((acc, curr) => {
            if (curr.name === 'name') {
              return true;
            } else {
              return acc;
            }
          }, false)
        );
      },```

File: client/Contexts/index.js [hunk 1]
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+export * from './Introspect/Context';
```
nictownsend: `.ts` instead?

pmuir:
This is ready for review - I've had to base it on top of @JQrdan 's GraphQL PR and that should merge first.

matthew-chirgwin:
@pmuir the client has now been merged - once rebased I will review this

client/Contexts/Introspect/Expectations.ts
```diff
@@ -0,0 +1,52 @@
...
+export const expectations = {
+    Topic: {
+        type: 'Topic', // the GraphQL type name
+        fields: {
+            name: 'String!', // the expected GraphQL type - could be string or object - which is checked if is available and thus query-able
+            partitions: 'Float',
+            replicas: 'Int',
+        },
+        operations: {
+            create: ({ mutations }) => {
```
nictownsend: Could possibly avoid duplication with a partial function:

```
const mutationValidator = (operationName) => ({mutations}) => mutations[operationName] !== undefined;
...
create : mutationValidator('createTopic')
```

client/Contexts/Introspect/Expectations.ts
```diff
@@ -0,0 +1,52 @@
...
+            delete: ({ mutations }) => {
+                return mutations['deleteTopic'] !== undefined;
+            },
+            findByName: ({ queries }) => {
+                return (
+                    queries['topicsByName'] !== undefined &&
+                    queries['topicsByName'].args
+                        .map((v) => (v.name === 'name' ? true : false))
+                        .reduce((acc, curr) => {
+                            if (curr) {
```
nictownsend: `return` will just exit the loop but not shortcut the reduce so a later false would result in accumulator having false.

I think you need:
```
return (curr || acc) ? true : false;
```

 You also need to seed the reduce with a default value in case there are no args - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Reduce_of_empty_array_with_no_initial_value

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,137 @@
...
+};
+
+/**
+ * Converts an array into an object, using the element of the array indexed with `by` as the key.
+ * @param array The array to convert
+ * @param by The index of the array to use as the key
+ * @return the object
+ */
+const indexBy = <T>(array: readonly T[], by: string) => {
+    return array.reduce((acc, curr) => {
```
nictownsend: Array of `T`, reduce into a single object hash whereby each `T` is stored using its value of `by` as the key in the hash - what if value of `by` is the same/missing for multiple items? And does TS handle cases where the value isn't a string?

Also I think referring to `index` is confusing - it's not the index of the array, it's the prop name to use as the key.

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @param by The index of the array to use as the key
+ * @return the fields of the type identified by thing, indexed using by
+ */
+const indexSchema = (
+    types: { [key: string]: IntrospectionType },
+    schema: IntrospectionSchema,
+    thing: string,
+    by: string
+) => {
+    const field = schema[thing];
```
nictownsend: This seems to be indexing a single "thing" from the schema, not the whole schema. Misleading function name?

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (!field || !field.name) {
+        return {};
+    }
+    const type = types[field.name];
+    if (type.kind !== 'OBJECT') {
```
nictownsend: What if the type can't be found? NPE as type is `undefined`

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,137 @@
...
+    schema: IntrospectionSchema
+): Entities => {
+    // index the types by name
+    const types = indexBy(schema.types, 'name');
+    return Object.fromEntries(
+        Object.entries(expectations).map(([entityName, metadata]) => {
+            const answer = {} as Entity;
+            // For each expectation, locate it in the schema
+            const type = types[metadata.type];
+            if (type.kind === 'OBJECT') {
```
nictownsend: NPE if type undefined

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+export const introspect = (
+    expectations: Expectations,
+    schema: IntrospectionSchema
+): Entities => {
+    // index the types by name
+    const types = indexBy(schema.types, 'name');
+    return Object.fromEntries(
+        Object.entries(expectations).map(([entityName, metadata]) => {
+            const answer = {} as Entity;
```
nictownsend: Do you need to define the shape with defaults? For e.g `answer.fields` may be left undefined below.

client/Contexts/Introspect/mock/expectations-basic.ts
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+import {Expectations} from "../ExpectationTypes";
```
nictownsend: Minor concerns that this duplicates a lot of the real `Expectations` - is it needed or can it be reduced to provide a smaller test case (e.g the reducers in the callbacks)

nictownsend (changes requested):

client/Contexts/Introspect/Expectations.ts
pmuir: To be clear, I don't expect this to be the *real* expectations for `Topic`, so it's probably better to actually just remove the contents.

client/Contexts/Introspect/introspection.ts
pmuir: > Array of T, reduce into a single object hash whereby each T is stored using its value of by as the key in the hash - what if value of by is the same/missing for multiple items?

I don't think it can be, in the usage of this function here (as the introspection schema won't return the same thing twice), but I'll add an explicit check and document it.

> And does TS handle cases where the value isn't a string?

I don't think it does so I've added an explicit check and documented it. Again, I don't believe in the usage of the function this error can occur (due to what the introspection schema returns).

> Also I think referring to index is confusing - it's not the index of the array, it's the prop name to use as the key.

Used your text.

client/Contexts/Introspect/introspection.ts
pmuir: Yeah, I really struggled with what to call this function. Suggestions?

client/Contexts/Introspect/mock/expectations-basic.ts
pmuir: Yeah, I removed the real Expectations - we haven't actually determined the real ones yet. This one purely exists for unit testing this code. We don't want to merge them as that would mean that changes to the real expectations incorrectly break tests.

client/Contexts/Introspect/introspection.ts
pmuir: Yes, I think that makes sense.

client/Contexts/Introspect/Expectations.ts
pmuir: How would it return false later?

pmuir: Anyway, I changed it to your form as that also works.

client/Contexts/Introspect/Expectations.ts
pmuir: I removed it.

dlabaj (approved):

client/Contexts/Introspect/mock/expectations-basic.ts
```diff
@@ -0,0 +1,64 @@
...
+      delete: ({ mutations }) => {
+        return mutations['deleteTopic'] !== undefined;
+      },
+      findByName: ({ queries }) => {
+        return (
+          queries['topicsByName'] !== undefined &&
+          queries['topicsByName'].args
+            .map((v) => (v.name === 'name' ? true : false))
+            .reduce((acc, curr) => {
+              return curr || acc ? true : false;
```
nictownsend: This comment depends on whether I've interpreted the intention of the code.

If this interpretation is correct:

```
- Iterate over the list of args, mapping to `true|false` based off whether the arg is called `"name"`
- Reduce that mapped array into a single boolean, which will be `true` if any of the args were called `name`
```

If that is the case - I think the original code had a `return false` if the current value was false. That meant that if the first arg mapped to `true`, but the remaining `n` mapped to `false` - the reduce result would be `false` (I don't think you can short circuit a `reduce` so the first `true` would later be overridden by the `false`s).

In addition; if my description is correct, you can simplify it with the following:

```
queries['topicsByName'].args.findIndex(v => v.name === 'name') > -1
```

nictownsend: I also note that this is now only in the mocks (but probably still valid in case this becomes a seed for the real expectations when later added back)

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,174 @@
...
+ * @param types The Types that exist on the schema
+ * @param schema The schema to index
+ * @param namedTypeRefKey The named type reference to use for indexing
+ * @param propName The prop name to use as the key
+ * @return the fields of the type identified by thing, indexed using by
+ * @throws Error If the types does not contain a type for the namedTypeRefKey
+ * @throws Error If the array contains more than one element that has the same value for by
+ * @throws Error If the key by is not of type string in the type identified by the namedTypeRefKey
+ */
+const indexSchema = (
```
nictownsend:
```suggestion
const indexTypeFromSchema = (
```
maybe? naming is hard :)

client/Contexts/Introspect/introspection.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw new Error(
+        `key identified by ${propName} in ${curr} must be of type "string"`
+      );
+    }
+    if (curr[propName] in acc) {
+      throw new Error(`key ${curr[propName]} already exists in ${acc}`);
+    }
+    acc[curr[propName]] = curr;
+    return acc;
+  }, {} as { [key: string]: T });
```
nictownsend: https://lodash.com/docs/4.17.15#keyBy - might simplify code with one caveat - it will override duplicate keys with latest value, rather than throw error (depending on whether that's important).

```suggestion
  return _keyBy(array, thing => {
  const prop = thing[propName];
    if (typeof prop !== 'string') {
      throw new Error(
        `key identified by ${propName} in ${curr} must be of type "string"`
      );
    }
    return prop;
```

nictownsend (changes requested):
These are pretty much semantic changes rather than blockers - so if you think it's not worth changing I'll change to approval.

client/Contexts/Introspect/introspection.ts
```diff
@@ -45,18 +45,14 @@ const transformFieldType = (type: IntrospectionTypeRef) => {
  * @throws Error If the key propName is not of type string in the array
  */
 const indexBy = <T>(array: readonly T[], propName: string) => {
-  return array.reduce((acc, curr) => {
-    if (typeof curr[propName] !== 'string') {
+  return _keyBy(array, thing => {
```
nictownsend: This will need an install of lodash's keyBy function package

nictownsend (changes requested):

nictownsend (approved):

package.json
```diff
@@ -49,21 +49,22 @@
...
+    "lodash.keyby": "^4.6.0",
     "lodash.merge": "^4.6.2",
     "mini-css-extract-plugin": "^1.2.1",
     "optimize-css-assets-webpack-plugin": "^5.0.4",
-    "react": "^16.13.1",
     "pino": "^6.7.0",
     "pino-filter": "^1.0.0",
     "pino-http": "^5.3.0",
     "pino-pretty": "^4.3.0",
-    "sass": "^1.29.0",
```
matthew-chirgwin: Looks like a rebase change - we were at 1.29, but after this would be a 1.26.10

client/Contexts/Introspect/mock/expectations-basic.ts
```diff
@@ -0,0 +1,60 @@
...
+      names: '[String]',
+      size: 'Int', // TODO how do we handle args?
+      roughSize: 'TShirtSize',
+      requiredRoughSize: 'TShirtSize!',
+      possibleRoughSizes: '[TShirtSize]',
+      requiredNames: '[String]!',
+      nonNullNames: '[String!]',
+      nonNullRequiredNames: '[String!]!',
+      topics: '[Topic]',
+    },
```
matthew-chirgwin, line 21: Is this just to illustrate a type can reference others? If so, would doc be a suitable place for it? Might save on the TODO too?

matthew-chirgwin (changes requested):
LGTM! Spotted a small (rebase) dependency change, and had one question around the `other` expectation defined

client/Contexts/Introspect/mock/expectations-basic.ts
pmuir: No, it's mock data used to test the introspector. It's important to validate that the instrospector works with all different types in GraphQL IMO.

package.json
pmuir: this is caused by mistake in the current `package.json` - see https://github.com/strimzi/strimzi-ui/blob/master/package.json#L66-L68 - `sass` is specified twice.

pmuir: I can fix it to use 1.29 though.

client/Contexts/Introspect/mock/expectations-basic.ts
pmuir: no, it's mock data used to validate the introspector

pmuir: The TODO is valid, but we can just make an issue for it - I don't think it's a common case.

client/Contexts/Introspect/Context.tsx
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import React, {ReactElement, ReactNode, useEffect, useState} from "react";
```
jeff-phillips-18: Prefer to just import `React` and prefix all the others rather than mixing.

client/Contexts/Introspect/Context.tsx
```diff
@@ -0,0 +1,93 @@
...
+ */
+export const IntrospectionContext = React.createContext({} as IntrospectionResult);
+
+/**
+ * IntrospectionProvider adds the introspection context.
+ * @param children
+ * @constructor
+ */
+export const IntrospectionProvider: React.FunctionComponent<IntrospectionProviderProps> = ({children}): ReactElement => {
+    const [result, setResult] = useState({
```
jeff-phillips-18:
```suggestion
    const [result, setResult] = useState<IntrospectionResult>({
```

client/Contexts/Introspect/Context.tsx
```diff
@@ -0,0 +1,93 @@
...
+    // The entities
+    entities: Entities
+    // True whilst the introspection result is loading
+    isLoading: boolean
+    // True when the introspection result is loaded
+    isComplete: boolean
+    // True if an error occurred whilst loading the introspection
+    isError: boolean
+    // The complete result from the server of the introspection query
+    response: ApolloQueryResult<IntrospectionQuery>
```
jeff-phillips-18:
```suggestion
    response: ApolloQueryResult<IntrospectionQuery> | null
```

client/Contexts/Introspect/Context.tsx
```diff
@@ -0,0 +1,93 @@
...
+            } as IntrospectionResult)
+        } catch (e) {
+            console.debug(`GraphQL Introspection Error ${e}`);
+            setResult({
+                entities: {},
+                isComplete: true,
+                isError: true,
+                isLoading: false,
+                response: {}
+            } as IntrospectionResult)
```
jeff-phillips-18:
```suggestion
                response: null
            })
```

client/Contexts/Introspect/Context.tsx
```diff
@@ -0,0 +1,93 @@
...
+    } as IntrospectionResult);
+
+    const fetchSchema = async () => {
+        setResult({
+            entities: {},
+            isComplete: false,
+            isError: false,
+            isLoading: true,
+            response: {}
+        } as IntrospectionResult)
```
jeff-phillips-18:
```suggestion
            response: null
        })
```

jeff-phillips-18 (approved):
Some suggestions, otherwise lgtm

pmuir:
I've pulled on the `Context` impl for now, will add it once #99 merges with the test utility for graphql fetching.

matthew-chirgwin (approved):
Changes LGTM - thanks :)

Wont block the PR on it, but could a brief summary of the files in the `mock` directory be added to the Introspect README in a following PR? Just to cover what they are, how to go about updating them (if required) etc?

scholzj (approved):

PR #59, opened 2020-10-29
feat: Add server-side logging capability

### LINKED ISSUE #43: Implement server-side logging
Issue to implement the server-side section of https://github.com/strimzi/strimzi-ui/pull/41

ajborley:
Contributes to: #43


server/core/app.ts
```diff
@@ -33,78 +34,50 @@ export const returnExpress: (
...
-      req.headers[STRIMZI_UI_REQUEST_ID_HEADER] = randomBytes(8).toString(
-        'hex'
-      );
-    }
-    const requestID = req.headers[STRIMZI_UI_REQUEST_ID_HEADER] as string;
-    // and make sure the response has the header
-    res.setHeader(STRIMZI_UI_REQUEST_ID_HEADER, requestID);
-    // create a 'context' for this request, containing config, a request ID, and loggers. Available to handlers via `res.locals.strimziuicontext`
+    // set up logger for start/end of request
+    httpLogger(req, res);
```
matthew-chirgwin: does the httpLogger also handle the end of the request?

matthew-chirgwin: To check: is this middleware? If so, should it be inserted as a part of the app.all?

matthew-chirgwin: To check: req.id contains the id, I assume generated in httplogger - should it not be in res.locals? (https://expressjs.com/en/api.html#res.local)

server/core/app.ts
```diff
@@ -33,78 +34,50 @@ export const returnExpress: (
...
-        'startup',
-        'core',
-        'moduleMount',
-        `Mounting module '${moduleName}'`
-      );
+      logger.info(`Mounting module '${moduleName}'`);
       const config = getConfig();
       const { mountPoint, routerForModule } = addModule(
-        (moduleName) => generateLoggers(serverName, moduleName, 'startup'),
+        (moduleName) => generateLogger(moduleName),
```
matthew-chirgwin: With the new logger here (and the fact we know the moduleName, server name etc), rather than pass the generator, could we pass the result of generateLogger's entry ( as we know the function name/params)? Addresses a concern @nictownsend had re the passing of the generator to all modules

server/logging.ts
```diff
@@ -0,0 +1,110 @@
...
+        loggerOptions,
+        err,
+      },
+      `Failed to configure logging options: ${err.message}`
+    );
+  }
+};
+
+/** Update the rootLogger when the config is loaded */
+loadConfig((loadedInitialConfig) => {
```
matthew-chirgwin: Rather than set up/load up/watch the config in multiple places (ie here and in main.ts), is there scope to have main.ts (when it loads/changes config) call a callback provided to do the same thing by the logger?

matthew-chirgwin (commented):
Looks good! Just a couple of questions/comments. Will also ask @nictownsend to review as he had thoughts around the log generator. 

Also adding @dlabaj / @pmuir to check this will work from their side

server/core/app.ts
ajborley: Yep - the httpLogger adds the end request logging by default. 
Yes pino-http can be treated as middleware - I will update to use it like that. 
And yep, req.id is added to the request in the pino-http code via the genReqID callback (https://github.com/ajborley/strimzi-ui/commit/52e636b23e35078c1d8c8bbed677d8e143501cac#diff-072fff4cca78a5c9ceb244882beffb3d8970dff40d5134e7c85c9a14b4e1548bR95) - not sure why it isn't added to res.locals - is that an express-only thing? pino-http can be applied to other web frameworks including express.

server/logging.ts
ajborley: Yep, will see what can be done.

server/core/app.ts
matthew-chirgwin: Sorry - forgot to come back and update these having read through everything!

I read from various sources that `res.locals` is the place to put per request state/data. I think they were of the opinion the request should be treated as immutable, but the response was ours to manipulate as we saw fit. Must admit I like that model, hence what I had previously. Thus my preference would be for it to be in res.locals, but happy to see what the consesnus is. All I would say is if we move it/add it, can we update the typings as needed: https://github.com/strimzi/strimzi-ui/blob/ae571ee62875127d08716a42beb1ba9f58f79651/server/types.ts#L143

server/core/app.ts
ajborley: Sure, will update.

server/core/app.ts
ajborley: The request ID is still being added to `res.locals.strimziuicontext.requestID` here: https://github.com/ajborley/strimzi-ui/commit/52e636b23e35078c1d8c8bbed677d8e143501cac#diff-7fde445974fe7c9eddd663ca12666ad12830c2800b3e1d8b3d731e8849870ceeR61
It's just generated via pino-http now and added (by pino-http) to the `req` object (which makes sense to me as the request ID should also be treated as immutable, alongside the rest of the request).

matthew-chirgwin (approved):
LGTM!

scholzj (approved):

package.json
```diff
@@ -44,6 +44,8 @@
   "dependencies": {
     "@babel/core": "^7.10.5",
     "@babel/preset-env": "^7.10.4",
+    "@types/pino": "^6.3.2",
+    "@types/pino-http": "^5.0.5",
```
pmuir: types normally go in `devDependencies`

pmuir (approved):

package.json
ajborley: Good spot! Have moved them in the latest commit 👍

package.json
```diff
@@ -42,6 +42,9 @@
   "bugs": {},
   "homepage": "https://github.com/strimzi/strimzi-ui#readme",
   "dependencies": {
+    "@babel/core": "^7.10.5",
```
matthew-chirgwin: With recent changes, babel should have been superceeded with TS. Where is this needed/used?

matthew-chirgwin (changes requested):
This will need a rebase/update as dependency changes now clash. I also have one question re babel which is being added (but not sure if now needed)

package.json
ajborley: I think this was a rebasing thing. Have rebased again & it's no longer there.

matthew-chirgwin (approved):
LGTM - thanks!

PR #74, opened 2020-11-02
feat: checkout correct ref on pr

### LINKED ISSUE #72: CI checks out master branch on PR instead of PR branch
With the move to `pull_request_target` as the github workflow trigger, it's now checking out master in the CI pipeline instead of the PR branch

JQrdan:
Closes: #72

### DIFF
--- a/.github/workflows/node-pr-jobs-secure.yml
+++ b/.github/workflows/node-pr-jobs-secure.yml
@@ -2,6 +2,7 @@
 # License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
 name: Node
 
+# Running target makes it secure as this is the only way to get hold of github secrets
 on: [pull_request_target]
 
 jobs:
--- a/.github/workflows/node-pr-jobs-secure.yml
+++ b/.github/workflows/node-pr-jobs-secure.yml
@@ -7,7 +7,7 @@ on: [pull_request_target]
 
 jobs:
   build-current:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: Build and Test
     steps:
       - name: Get PR head
--- a/.github/workflows/node-pr-jobs.yml
+++ b/.github/workflows/node-pr-jobs.yml
@@ -6,7 +6,7 @@ on: [pull_request]
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: Linter
     steps:
       - uses: actions/checkout@v2
@@ -20,7 +20,7 @@ jobs:
         run: npm run lint
 
   e2e-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: E2E Tests
     steps:
       - uses: actions/checkout@v2
@@ -38,7 +38,7 @@ jobs:
           path: failure_output
 
   storybook-artifact:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: PR Storybook
 
     steps:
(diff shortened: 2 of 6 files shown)


nictownsend:
LGTM and https://github.com/JQrdan/strimzi-ui/pull/5 shows the action that runs on base repo as working

nictownsend (approved):

.github/workflows/node-pr-jobs-secure.yml
```diff
@@ -0,0 +1,68 @@
...
+
+jobs:
+  build-current:
+    runs-on: ubuntu-latest
+    name: Build and Test
+    steps:
+      - name: Get PR head
+        id: pr_head
+        run: |
+          OWNER=$(jq -r '.pull_request.base.repo.owner.login' "$GITHUB_EVENT_PATH")
```
matthew-chirgwin, line 16: Is it a guarantee we will always have jq from `ubuntu-latest`?

.github/workflows/node-pr-jobs.yml
```diff
@@ -2,55 +2,9 @@
 # License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
 name: Node
 
-on: [pull_request_target]
+on: [pull_request]
```
matthew-chirgwin, line 5: I assume this is seperate so `pull_request` refers to ref/branch being PR'ed - ie no need to HEAD as per above? I assume there is no way that information could be discovered once and shared between jobs (so we only have one workflow)?

matthew-chirgwin (commented):
Looks good - and good spot in the first instance! Have some questions about the change, but none I would reject the PR on. Will also ask for a review from the maintainers/Pete/Don

pmuir (approved):
I've not used github actions much, generally 👍

.github/workflows/node-pr-jobs-secure.yml
JQrdan: Good point, might be worth changing these to an LTS release

.github/workflows/node-pr-jobs.yml
JQrdan: Yep! I don't see the harm in these ones being normal `pull_request` workflow triggers as they don't need any secrets. It _might_ be possible to share the pr_head info as an artefact but given my previous fun with artefact downloading, I thought it would be best to leave that for now.

JQrdan:
> I've not used github actions much, generally 👍

We've only played around with them slightly in the past but really liked them. Hence the teething problems with different GitHub topologies :D

matthew-chirgwin (changes requested):
> Good point, might be worth changing these to an LTS release

Any plans to make a change to do this?

> Yep! I don't see the harm in these ones being normal `pull_request` workflow triggers as they don't need any secrets. It _might_ be possible to share the pr_head info as an artefact but given my previous fun with artefact downloading, I thought it would be best to leave that for now.

In a perfect world, we would have just the one file. I do wonder if two files will scale/last, but will not ask for early optimisation without a good use case. As a blocker however - it looks like e2es, storybook and linting are run in both `pull_request_target` and `pull_request` events. If the jobs are running repeatedly, can they be only run on one? There is a limit on the number of actions that can run at one time across the whole org I believe - we do not want to be a bad actor and waste resources.

scholzj (approved):

JQrdan:
> In a perfect world, we would have just the one file. I do wonder if two files will scale/last, but will not ask for early optimisation without a good use case. As a blocker however - it looks like e2es, storybook and linting are run in both `pull_request_target` and `pull_request` events. If the jobs are running repeatedly, can they be only run on one? There is a limit on the number of actions that can run at one time across the whole org I believe - we do not want to be a bad actor and waste resources.

@matthew-chirgwin these are being duplicated in this PR as they have been defined as `pull_request` in this PR but are `pull_request_target` in master so there's a clash. Once merged, that will go away

matthew-chirgwin (approved):

PR #79, opened 2020-11-03
Upgrade to webpack 5

### LINKED ISSUE #54: Enable Module Federation (Upgrade to Webpack 5)
We intend to use [Webpack Module Federation](https://webpack.js.org/concepts/module-federation/) to build Red Hat product UIs from Strimzi UI.

In order to do this we will need to enable Module Federation in Strimzi UI, which will involve:

* Add this to the relevant architecture documentation
* Upgrading Webpack to the latest v5 release (Module Federation was added in Webpack 5)
* Upgrading all Webpack loaders & plugins to the latest version (this will ensure they are compatible with Webpack 5)
* Add the module federation plugin to the webpack config, and declare source files we want to expose
* Add shared dependencies (`react`, `react-dom`) as shared dependencies
* Add default exports to each source file we want to expose (with the default export pointing at the component we are exporting)

This also has an impact on how we design our React components as we need to start designing them for reuse in unknown situations, not just use within in the strimzi-ui. Designing components for reuse via module federation is no different to designing them for reuse via npm dependency management.

A few component design rules that will make reuse much easier are:

* create "pages" that wrap components. For each view that we want to create, build the component separately from the page, allowing the the component to be reused. A page might have titles, padding, layout etc. The component is the actual interaction.
* create "dumb" components that are wrapped by "smart" components (this is also very useful for mocking data for testing, storybook etc.). Here for each component we create we pass in any data that is needed using props, rather than directly accessing it. This allows the data sources to be altered in various ways by the reusing UI. Smart components are often pages as well.
* use callbacks to handle transitions to/from views in dumb components. In other words don't navigate using `history` but instead pass a callback in as a prop. This allows the component to be reused in a different setting.
* make components more granular. Don't build the entire view in a single component - split it up to allow reuse.

pmuir:
* move all webpack dependencies and babel to devDependencies
* upgrade all webpack dependencies to latest version
* switch to webpack serve
* switch to css-mimimizer-webpack-plugin from optimize-css-assets-webpack-plugin as suggested by optimize-css-assets-webpack-plugin maintainers

See #54 


pmuir:
This is the big changes required for webpack 5 & module federation. I'll add the federation plugin in a subsequent PR with the aim of not letting PRs get too large.

nictownsend:
@pmuir - what are your thoughts over `deps` vs `devDeps` - @matthew-chirgwin and I have previously pivoted from `deps` meaning "only things you need to run the project after building", to also include "anything that creates build output".

Argument being - automated scans that check for vulnerable libraries need to flag things that could be injecting code into the build (webpack for e.g) - but they don't need to run against things like jest or helper modules for testing.

nictownsend (changes requested):
Overall LGTM - but looks like webpack is not happy with storybook, and would be good to get thoughts on my comment above too, thanks!

pmuir:
AIUI `deps` is only for production dependencies - see https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file. With things like this I would tend to follow the official advice.

I think that if the tool isn't scanning your development time dependencies and you want it to, you probably want to configure the tool, not deviate from standard packaging.

pmuir:
And yes, storybook build is broken - see issue.

pmuir:
@b1zzu managed to find an upgraded html-webpack-plugin that will make this work.

nictownsend:
> AIUI `deps` is only for production dependencies - see https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file. With things like this I would tend to follow the official advice.
> 
> I think that if the tool isn't scanning your development time dependencies and you want it to, you probably want to configure the tool, not deviate from standard packaging.

@matthew-chirgwin I'm inclined to agree (with assumption that whatever tooling is picked can be flagged to add certain dev tooling to an allow list)

pmuir:
@matthew-chirgwin can we get this merged?

matthew-chirgwin:
> > AIUI `deps` is only for production dependencies - see https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file. With things like this I would tend to follow the official advice.
> > I think that if the tool isn't scanning your development time dependencies and you want it to, you probably want to configure the tool, not deviate from standard packaging.
> 
> @matthew-chirgwin I'm inclined to agree (with assumption that whatever tooling is picked can be flagged to add certain dev tooling to an allow list)

@pmuir / @nictownsend - re dependencies, the issue for me is what we are producing is not an npm package we will upload to npm, and moreover, the definition of production and development dependancies do not fit what we are producing either. Hence the initial write up here: https://github.com/strimzi/strimzi-ui/blob/master/docs/Architecture.md#dependency-management

If we were producing a package which would be published to NPM, we would and should follow the convention - it would be mad not to, as that is what consumers will expect. However, we are not producing an NPM package in this repo. Rather, we are building a UI for Strimzi, which will be distributed via Dockerfile so it can be deployed by the Strimzi operator. Thus, really we are a consumer of the NPM registry, and thus, all we would look to do is pull dependencies we need to make the UI. For me, there then is a question: how should we classify those dependencies?

In my opinion, the npm dependency vs dev dependency definition (originally intended for node modules/packages) does not work/has not aged well for more modern web UIs. `Packages required by your application in production.` - when is `production` for a built UI? A good number of our 3rd party dependancies effect/are used in the ‘built’ code - ie they either take what we write and either compile, transform, or add their own inserts to in addition/based on what we write. 

Hypothetically, if we use a webpack plugin (for example) as a dev dependency (the traditional location for npm published packages), and it inserts a polyfill to our built bundle, we would ship it as a part of our built UI output. If that polyfill is somehow compromised, and malicious code inserted, that malicious code is then shipped as well. This would mean the malicious code could then effect consumers/users (either as consumers of an NPM package, or in our case, with malicious code running in the user’s browser or on the UI server).

npm’s own definition of a dependency is `Packages required by your application in production.`. If it is built, or results in built output, for me this makes it `required` in production. Hence the guidance in the link above.

These transpiling/compiling/polyfilling type libraries are in an interesting grey area - one for which the normal dependency vs dev dependency classification doesn’t really fit. Peer is not right either - as a consumer may not need to use the same build stack as us. It also doesn’t help either differentiating these libraries and what they do is non trivial (ie webpack is the runtime, and doesn’t modify code, but webpack includes plugins which we use/does). The general answer I have seen elsewhere in these cases is list everything as a dependency - as you need them all to develop the codebase in question. I err on that, as there is still a useful differentiation to make between what we ship, and what we use to develop.

With sensible differentiation of dependencies, we can still get value from production and development classified packages. For one, at build time, we can install and include just the dependencies we need [in the dockerfile at runtime.](https://github.com/strimzi/strimzi-ui/blob/26898902e404b4a4736f2fae3b106a63218bf5bc/build/dockerfile#L18) _Note: this will include build dependancies as well, but as per the above without another classification, this is a scenario I think is preferable to ship a few extra things which are never refernced in the dockerfile vs the whole set of dependencies._ In addition, by clarifying dependencies sensibly, we get value from tools like `npm audit`, which will flag issues with the packages we use directly or indirectly, and in our case, where these issues would manifest - ie dev only, production only, both etc.

In short, without further discussion/agreement  to change strategy, I would reject this PR on the grounds a few dependencies are not in the right section as is today, with the reasons for those classifications as is hopefully described here. Really this is a wider issue, one that npm should guide on. I suspect for legacy reasons however, things would not change. Otherwise @pmuir I have no objection/issue with the change.

pmuir:
I can change them back to what they were before - it's an weird layout but whatever, it doesn't really matter.

pmuir:
@matthew-chirgwin ok - dependencies are now just upgraded.

package.json
```diff
@@ -146,8 +147,8 @@
     "ts-node": "^9.0.0",
     "tsconfig-paths": "^3.9.0",
     "tsconfig-paths-webpack-plugin": "^3.3.0",
-    "webpack-bundle-analyzer": "^3.8.0",
-    "webpack-cli": "^3.3.12",
+    "webpack-bundle-analyzer": "^3.9.0",
+    "webpack-cli": "^4.1.0",
     "webpack-dev-server": "^3.11.0",
```
matthew-chirgwin, line 152: Wont block the PR on it, but as `webpack serve` is being used, is dev server still needed? Perhaps worth removing in the next PR if so?

matthew-chirgwin (approved):

pmuir:
Yes, it's still needed. They only changed the command, not the dependencies.

scholzj (approved):

PR #82, opened 2020-11-04
fix: CI now pulls from forked repo

### LINKED ISSUE #72: CI checks out master branch on PR instead of PR branch

JQrdan:
Closes: #72

### DIFF
--- a/.github/workflows/node-pr-jobs-secure.yml
+++ b/.github/workflows/node-pr-jobs-secure.yml
@@ -10,17 +10,6 @@ jobs:
     runs-on: ubuntu-20.04
     name: Build and Test
     steps:
-      - name: Get PR head
-        id: pr_head
-        run: |
-          LABEL=$(jq -r '.pull_request.head.label' "$GITHUB_EVENT_PATH")
-          REPO=$(jq -r '.pull_request.base.repo.name' "$GITHUB_EVENT_PATH")
-          OWNER=${LABEL%:*}
-          OWNER+=/
-          PR_REPO="$OWNER$REPO"
-          PR_BRANCH=$(jq -r '.pull_request.head.ref' "$GITHUB_EVENT_PATH")
-          echo "::set-output name=repo::$PR_REPO"
-          echo "::set-output name=branch::$PR_BRANCH"
       - name: Checkout master
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


JQrdan:
Verified here: https://github.com/JQrdan/strimzi-ui/pull/9

matthew-chirgwin (commented):
I helped test the change, and it did work on @JQrdan 's fork from mine. I am happy with the change, but given the churn, welcome additional reviews

.github/workflows/node-pr-jobs-secure.yml
```diff
@@ -28,10 +35,10 @@ jobs:
           echo "SERVER_MASTER_REPORT=$(cat generated/bundle-analyser/server-report.json)" >> $GITHUB_ENV 
       - name: Checkout PR Branch
         uses: actions/checkout@v2
-        # with:
-        #   repository: ${{ steps.pr_head.outputs.repo }}
-        #   ref: ${{steps.pr_head.outputs.branch}}
-        #   fetch-depth: 0
+        with:
+          repository: ${{ steps.pr_head.outputs.repo }}
+          ref: ${{steps.pr_head.outputs.branch}}
```
nictownsend: https://github.community/t/running-code-from-forks-with-pull-request-target/126688/6 - does that work? If so, seems nicer than having to `jq` and export the ref.

Also, earlier in that thread it does mention risk of exporting tokens (i.e I write a script/action that can export these secrets) - not sure if that's a worry?

nictownsend (commented):
Left comment re approach. If it's not feasible then I'll approve.

scholzj (approved):

.github/workflows/node-pr-jobs-secure.yml
JQrdan: Hmm good find, I'll give that a go now

The initial problem came from token security so this was supposed to work with that. Actions will mask any sensitive data and only runs because it knows it trusts the source

JQrdan: That did the trick, I missed the `full_name` key which basically solves all of the problems I had 😂

nictownsend (approved):

pmuir (approved):

PR #85, opened 2020-11-05
feat: Add client-side logging support

### LINKED ISSUE #42: Implement logging in client
Issue to implement the client-side section of https://github.com/strimzi/strimzi-ui/pull/41

ajborley:
- Adds the websocket listener to the server.
- Adds support on the client to send logging events through the
websocket via a new `useLogger` hook.

Contributes to: #42


ajborley:
Draft PR to check if folks think I'm on the right track... still working on how to test this all.

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,94 @@
...
+  clientID: string;
+  clientLevel: LogLevel;
+  componentName: string;
+  msg: string;
+}
+
+export const useLogger: (string) => (LogLevel, string) => void = (
+  componentName: string
+) => {
+  // TODO: store the logger websocket in context, so you get one for the whole UI, rather than one every time you call useLogger.
```
pmuir: Feels like we need this before we can merge?

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,94 @@
...
+          loggerMessageBuffer.splice(
+            0,
+            loggerMessageBuffer.length - MESSAGE_BUFFER_MAX_SIZE
+          );
+        }
+      }
+    }
+  };
+
+  return logger;
```
nictownsend: Could we have the standard `log` `debug` etc as curried methods instead of passing the level in whenever you call the logger method?

```
const genLog = (logLevel) => (message) => {
  ...
 buffer.push({logLevel, message});
  ...
};


return {
  info : genLog("info);
  ...
}
```

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,94 @@
...
+  const logger = (clientLevel: LogLevel, msg: string) => {
+    // Check if the component should be logged
+    if (defaultLoggingRegex.test(componentName)) {
+      const loggerWebSocket = loggerWebSocketRef.current;
+      const loggerMessageBuffer = loggerMessageBufferRef.current;
+
+      // Add a new new message to the buffer
+      loggerMessageBuffer.push({
+        clientTime: Date.now(), // client timestamp in ms
+        clientID: defaultClientID, // client ID
```
nictownsend: UUID? Or something sent up from server when the socket opens?

server/log/router.ts
```diff
@@ -14,8 +21,51 @@ export const LogModule: UIServerModule = {
...
 
     return exit({ mountPoint: '/log', routerForModule });
   },
 };
+
+const handleMessage: (logger: Logger) => (data: Data) => void = (logger) => (
+  data
+) => {
+  if (typeof data === 'string') {
+    try {
```
nictownsend, line 44: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c1c23b2dcaf1a419992379a2b5bf7e14fe829f7a/types/ws/index.d.ts#L118 - any chance we could end up with buffers? Is it client that decides the format?

server/log/router.ts
```diff
@@ -14,8 +21,51 @@ export const LogModule: UIServerModule = {
...
+      JSON.parse(data).forEach((clientLogEvent: ClientLoggingEvent) => {
+        if (clientLogEvent.clientLevel) {
+          logger[clientLogEvent.clientLevel](
+            clientLogEvent,
+            clientLogEvent.msg
+          );
+        } else {
+          logger.debug(clientLogEvent, clientLogEvent.msg);
```
nictownsend: https://github.com/pinojs/pino/blob/HEAD/docs/api.md#logging-method-parameters - looks like we don't need second param as our event contains `msg` field

server/log/router.ts
```diff
@@ -14,8 +21,51 @@ export const LogModule: UIServerModule = {
...
+  } else {
+    // Ignore any non-string data
+    logger.trace(`handleMessage ignoring data of type ${typeof data}`);
+  }
+};
+
+const handleClose: (
+  logger: Logger
+) => (code: number, reason: string) => void = (logger) => (code, reason) =>
+  logger.debug(`WebSocket listener closed. (${code}) ${reason}`);
```
nictownsend, line 67: Can we log out a request ID / client ID?

server/log/router.ts
```diff
@@ -14,8 +21,51 @@ export const LogModule: UIServerModule = {
...
+        // Return 426 Upgrade Required if this isn't a websocket request
+        res.sendStatus(426);
+      }
+    });
 
     return exit({ mountPoint: '/log', routerForModule });
   },
 };
+
+const handleMessage: (logger: Logger) => (data: Data) => void = (logger) => (
```
nictownsend: Pedantic, but can you rename to `messageHandler` (as it's not itself handling it, but the return function.

server/log/router.ts
```diff
@@ -14,8 +21,51 @@ export const LogModule: UIServerModule = {
...
     // implementation to follow
-    routerForModule.get('*', authFn, (req, res) => res.sendStatus(200));
+    routerForModule.get('*', authFn, (req, res) => {
+      const { isWs } = req as strimziUIRequestType;
+      const { ws } = res as strimziUIResponseType;
+      if (isWs) {
+        ws.on('message', handleMessage(logger));
+        ws.on('close', handleClose(logger));
+      } else {
+        // Return 426 Upgrade Required if this isn't a websocket request
```
nictownsend, line 31: Is this just belts and braces on the error case? Could return `404` as it's not a valid call for non web?

server/main.ts
```diff
@@ -69,6 +70,18 @@ loadConfig((loadedInitialConfig) => {
...
 
+  const wss = new WebSocket.Server({ noServer: true });
+  instance.on('upgrade', (req, socket, head) =>
+    wss.handleUpgrade(req, socket, head, (ws) => {
+      const res = new http.ServerResponse(req) as strimziUIResponseType;
+      // add/mark the request as a websocket request
+      req.isWs = true;
+      res.ws = ws;
+      // call the express app as usual
+      expressApp(req, res);
```
nictownsend: How does `expressApp` know which endpoint to call? Is it simply that we're listening on `host:port/*` so any websocket request will be upgraded, but then sending messages will fail unless it's on a path that has `isWs` support?

server/test_common/commonServerSteps.ts
```diff
@@ -83,4 +119,83 @@ When(
...
+    websocket.close(0, 'Test close');
+  })
+);
+
+And(
+  'the WebSocket is closed',
+  stepWithWorld(async (world) => {
+    const { websocket } = world;
+    if (websocket.onclose != null) {
+      expect((<jest.Mock>websocket.onclose).mock.calls.length).toBe(1);
```
nictownsend: `toHaveBeenCalled` and related - https://jestjs.io/docs/en/expect#tohavebeencalled - is a neater API than `mock.calls.x`

nictownsend (changes requested):

client/Hooks/useLogger/Hook.ts
ajborley: Good idea - done.

client/Hooks/useLogger/Hook.ts
ajborley: Yep - have picked up the `uuid` package.

server/log/router.ts
ajborley: Yep - the `useLogger` hook will always send strings so this should be fine.

server/log/router.ts
ajborley: Good spot, have removed.

server/log/router.ts
ajborley: The close event only receives the code and reason string: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ws/index.d.ts#L86 so we would need to encode the client ID in the reason string. I will look into this when I hook into the context, as we may want to explicitly close the logging websocket on client unload.

server/log/router.ts
ajborley: Done

server/log/router.ts
ajborley: We expect HTTP GET calls to the `/log` endpoint to be websocket upgrade requests. If we get a request that is not an upgrade, then we return the [426](https://httpstatuses.com/426) response, as we don't serve any other http traffic from `/log`.

server/main.ts
ajborley: Yep, that's right. The code in main.ts (https://github.com/ajborley/strimzi-ui/blob/42-client-side-logging/server/main.ts#L74-L90) watches for websocket upgrade requests, and adds the `isWs` boolean and `ws` websocket to the req and res respectively. So currently, any endpoint that receives an upgrade request will set up a websocket, but only modules that implement the `websocket.onmessage`, etc handlers will actually do anything.

I expect this is not what we want - we should probably add something to the module configuration that can be used to decide whether the websocket upgrade request is permitted or not. What do you think?

server/test_common/commonServerSteps.ts
ajborley: Forgot all about the `toHaveBeenCalled*` API.  :) Updated

client/Hooks/useLogger/Hook.ts
ajborley: Agreed, and the same for the other TODOs (https://github.com/strimzi/strimzi-ui/pull/85/files#diff-3e52b502269a41a2c06ce466fd7d975b726b3557bc98a54d84b74b2bf958f1baR9-R12)

server/main.ts
nictownsend: I wonder if we need a break out session on this.

It seems like we're possibly shoehorning in the `log` module into express, when actually this is a different paradigm. My concern is that by appending `ws` to `req` we lose reference to it outside of the initial request.

Without extending express or using old express middleware - it feels like we need the ability to control "websocket" modules too - maybe, by having the concept of attaching multiple endpoints (https://www.npmjs.com/package/ws#multiple-servers-sharing-a-single-https-server) - we then destroy socket if it's not on a path that's acceptable?

We can then have "express" modules and "ws" modules (with a similar `addModule` for ws that takes in a logger)

matthew-chirgwin:
Requesting (re) reviews please @nictownsend / @pmuir !

client/Hooks/useLogger/README.md
```diff
@@ -0,0 +1,45 @@
...
+    }, 5000);
+  }, []);
+
+  return (<div {...props}/>);
+};
+```
+
+## API
+
+### `userLogger(componentName: string)`
```
nictownsend:
```suggestion
### `useLogger(componentName: string)`
```

client/Hooks/useLogger/Hook.ts
nictownsend: Is this still blocking merge? Feels like yes?

client/Hooks/useLogger/useLogger.steps.tsx
```diff
@@ -0,0 +1,131 @@
...
+let rerenderHook: (newProps?: unknown) => void;
+let mockWebSocketServer: Server;
+let onConnectionPromise: Promise<void>;
+let onMessagePromise: Promise<string>;
+let sentMessagesCount = 0;
+let verifiedClient = true;
+
+After(() => {
+  sentMessagesCount = 0;
+  mockWebSocketServer.close();
```
nictownsend: May be undefined if the `Given` was never called

client/Hooks/useLogger/useLogger.steps.tsx
```diff
@@ -0,0 +1,131 @@
...
+  verifiedClient = true;
+});
+
+Given('a logging WebSocket server', () => {
```
nictownsend: Worry is that if jest runs tests in parallel, or someone else uses this `Given` we'll get port clashes. But if this is fixed via the context change as a later commit (or new PR) I'm less worried. Wonder if when you add in the `context` we can supply a mock ws client via context instead of having to start up real servers in the test.

client/Hooks/useLogger/useLogger.steps.tsx
```diff
@@ -0,0 +1,131 @@
...
+  for (let index = 0; index < expectedLoggingMessagesCount; index++) {
+    expect(parsedLoggingMessages[index].clientLevel).toEqual(logLevel);
+    expect(parsedLoggingMessages[index].componentName).toEqual('TestComponent');
+    expect(parsedLoggingMessages[index].msg).toEqual(
+      `useLogger test message ${index + startingIndex}`
+    );
+  }
+};
+Then(loggingMessagesAreReceived, loggingMessagesAreReceivedFn);
+And(loggingMessagesAreReceived, loggingMessagesAreReceivedFn);
```
nictownsend, line 157: Eugh, that's unfortunate syntax :(

server/main.ts
```diff
@@ -69,6 +70,26 @@ loadConfig((loadedInitialConfig) => {
...
+    );
+
+    if (upgrade === 'websocket') {
+      wss.handleUpgrade(req, socket, head, (ws) => {
+        const res = new http.ServerResponse(req) as strimziUIResponseType;
+        // add/mark the request as a websocket request
+        req.isWs = true;
+        res.ws = ws;
+        // call the express app as usual
+        expressApp(req, res);
```
nictownsend, line 88: Please see previous comment re breakout - not sure how safe this is for GC purposes that we lose reference to the ws (to be able to close it etc)

nictownsend (changes requested):

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export type LogLevelType =
+  | 'fatal'
+  | 'error'
+  | 'warn'
+  | 'info'
+  | 'debug'
+  | 'trace';
+
+interface loggerMessage {
```
dlabaj: Are these all required?

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,125 @@
...
+  componentName: string
+) => {
+  // TODO: store the logger websocket in context, so you get one for the whole UI, rather than one every time you call useLogger.
+  const loggerWebSocketRef = useRef<WebSocket | null>(null);
+  const loggerMessageBufferRef = useRef<loggerMessage[]>([]);
+
+  useEffect(() => {
+    // Check if the component should be logged
+    if (defaultLoggingRegex.test(componentName)) {
+      const loggerWebSocket = loggerWebSocketRef.current;
```
dlabaj: Did we want to declare these types instead of using inference?

dlabaj (approved):
A few comments otherwise looks good.

client/Hooks/useLogger/Hook.ts
pmuir: I think it does. Opening a websocket for every log request defeats the purpose of using a websocket.

pmuir (changes requested):
TODOs need resolving, otherwise LGTM

server/main.ts
matthew-chirgwin: @ajborley / @nictownsend has this chat happened yet? Would be keen to participate. If the `ws` was placed in `res.locals`, it will be scoped to the life of the request (ie socket lifetime). Express will then discard the `locals`, `ws` included. Thoughts?

matthew-chirgwin (commented):
LGTM - re the TODOs, the first stage of config work has now landed - would it be possible to get the various TODOs re config usage to be updated to use that, which I can then build upon in #99 ?

client/Hooks/useLogger/README.md
ajborley: Fixed

client/Hooks/useLogger/useLogger.steps.tsx
ajborley: Check added

client/Hooks/useLogger/useLogger.steps.tsx
ajborley: The `new Server(..)` line here is a [`mock-socket`](https://www.npmjs.com/package/mock-socket) server, which just hooks mock-clients to mock-servers using the supplied server URL as a key. No real servers are being started here :)

client/Hooks/useLogger/useLogger.steps.tsx
ajborley: Agreed, ugly. Couldn't spot any alternative way to use `And` for the same code as a `Given`, `When` or `Then`. If you have any good ideas... :)

server/main.ts
ajborley: I think this will be superseded when Matt's websocket handling changes go in, so I've left it as-is for now.

client/Hooks/useLogger/Hook.ts
ajborley: These are all the levels available in the `pino` library used on the server. Only those at the `pino` logging level will actually get emitted to the log.

client/Hooks/useLogger/Hook.ts
ajborley: This code has been updated, hopefully the types are clearer now.

client/Contexts/Logging/Logging.types.ts
```diff
@@ -0,0 +1,24 @@
...
+  msg: string;
+}
+
+export type LogLevelType =
+  | 'fatal'
+  | 'error'
+  | 'warn'
+  | 'info'
+  | 'debug'
+  | 'trace';
```
JQrdan: Would this be nicer as an enum?

client/Contexts/Logging/README.md
```diff
@@ -0,0 +1,3 @@
+# Logging
+
+This context will store the logging WebSocket cient connection and the buffer of logging messages used in the (useLogger)[../../Hooks/useLogger] hook. This allows the useLogger hook to use a single WebSocket connection for all logging calls.
```
JQrdan:
```suggestion
This context will store the logging WebSocket client connection and the buffer of logging messages used in the (useLogger)[../../Hooks/useLogger] hook. This allows the useLogger hook to use a single WebSocket connection for all logging calls.
```

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Checks if the component should be logged by checking if the LOGGING
+ * query parameter regex metches the component name.
+ *
+ * @param componentName the component name to check
+ */
+const shouldLogComponent: (componentName: string) => boolean = (
+  componentName: string
+) => {
```
JQrdan: You can streamline typing like this:

```suggestion
const shouldLogComponent = (
  componentName: string
): boolean => {
```

then `shouldLogComponent` will infer the type from the right side. It saves having to write parameters and their types twice, which becomes painful when you change them

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (loggingParam) {
+      const loggingRegex = new RegExp(loggingParam);
+      shouldLogComponent = loggingRegex.test(componentName);
+    }
+  }
+
+  return shouldLogComponent;
+};
+
+const getWebSocketAddress: () => string = () => {
```
JQrdan:
```suggestion
const getWebSocketAddress = (): string => {
```

client/Hooks/useLogger/Hook.ts
```diff
@@ -0,0 +1,132 @@
...
+  return `${window.location.protocol === 'https:' ? 'wss' : 'ws'}://${
+    window.location.host
+  }/log`;
+};
+
+export const useLogger: (componentName: string) => LoggerType = (
+  componentName: string
+) => {
```
JQrdan:
```suggestion
export const useLogger = (
  componentName: string
): LoggerType => {
```

client/Hooks/useLogger/useLogger.steps.tsx
```diff
@@ -0,0 +1,167 @@
...
+  sentMessagesCount = 0;
+  if (mockWebSocketServer) {
+    mockWebSocketServer.close();
+  }
+  window.location = originalWindowLocation;
+});
+
+Given(/^a(?: (secure))* logging WebSocket server$/, (isSecure) => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
JQrdan, line 52: Not a review comment but just wanted to say this is amusing 😂

client/Hooks/useLogger/useLogger.steps.tsx
```diff
@@ -0,0 +1,167 @@
...
+const useLoggerHookIsRenderedFn = () => {
+  const initialRef = { current: { websocket: null, messageBuffer: [] } };
+  (useRef as jest.Mock).mockReturnValue(initialRef);
+  const wrapper = ({ children }) => (
+    <LoggingProvider>{children}</LoggingProvider>
+  );
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const renderedHook = renderHook(() => useLogger('TestComponent'), {
+    wrapper,
+  });
```
JQrdan, line 87: I'm curious why you need this eslint comment? I though renderHook behaved so that it should abide by all the rules

JQrdan (changes requested):

client/Hooks/useLogger/Hook.ts
ajborley: Latest commit resolves this via context

client/Contexts/Logging/Logging.types.ts
ajborley: Yep, have updated.

client/Contexts/Logging/README.md
ajborley: Done

client/Hooks/useLogger/Hook.ts
ajborley: Nice! Thanks for that, makes it much tidier :)

client/Hooks/useLogger/Hook.ts
ajborley: Done

client/Hooks/useLogger/Hook.ts
ajborley: Done

client/Hooks/useLogger/useLogger.steps.tsx
ajborley: I know :)

client/Hooks/useLogger/useLogger.steps.tsx
ajborley: Without it the `renderHook` line fails the linter, even though I'm following the [renderHook examples](https://react-hooks-testing-library.com/usage/basic-hooks) (I'm guessing the lint rules just aren't smart enough yet):
```
/strimzi-ui/client/Hooks/useLogger/useLogger.steps.tsx
  84:41  error  React Hook "useLogger" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function  (react-hooks/rules-of-hooks)
```

JQrdan: Very strange, in my code I have `const { result } = renderHook(() => useRouteConfig(input));` and it's happy

JQrdan (approved):
LGTM

server/main.ts
matthew-chirgwin: Yep - I will update the approach in my next PR

matthew-chirgwin (approved):

scholzj (approved):

PR #92, opened 2020-11-06
fix: split jest runs into client and server

nictownsend:
- allows v8 coverage provider to function with correct coverage

Contributes to: mhub/qp-planning#91


ajborley (approved):

dlabaj (approved):
LGTM

JQrdan:
I'll run these changes against my branch to double check

nictownsend:
For context - from what I can see, with the default (`babel`) provider, it seems to be using the configuration from the last jest project (which is `server` and such the `tsconfig` doesn't have jsx flag or module paths for `client`).

Switching to `v8` seems to use the `tsconfig` correctly from the `client`/`server` folders accordingly.

nictownsend:
Found out that coverage is unreliable when running a subset of tests (or single test) with v8.

Also github was providing different answers to local.

Bit the bullet and configured two jest runs - one for server, one for test - as they need different tsconfig (and coverage is not supporting in on jest). 

Will raise a jest issue (if I can make minimal repro)

matthew-chirgwin (approved):
LGTM

JQrdan (approved):
LGTM

scholzj (approved):

Issue #68, opened 2020-10-30
Navigation Component


matthew-chirgwin:
Issue to implement the Navigation Bootstrap component described in https://github.com/strimzi/strimzi-ui/blob/master/docs/Architecture.md#routing-and-navigation / https://github.com/strimzi/strimzi-ui/blob/master/client/Bootstrap/Navigation/README.md

- Requires Introspection, Feature flagging and testing issues to be completed

JQrdan:
This most certainly has seen activity

PR #93, opened 2020-11-09
feat: add useRouteConfig hook

JQrdan:
Doesn't include any additional checks such as feature flag checks,
auth checks, and backend checks yet

Contributes to: #68


dlabaj (approved):

.github/workflows/node-pr-jobs-secure.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       - name: Test client code
         run: npm run test:jest:client -- --forceExit # force exit in case a test does not clean up after failure
       - name: Test server code
+        if: ${{ always() }}
```
matthew-chirgwin: Run these even if a prior step fails?

client/Hooks/README.md
```diff
@@ -26,4 +26,4 @@ Where:
 
 ## Implemented hooks
 
-Hooks to be added here on implementation, with summary of purpose/usage and a link to it's README.
+- [useRouteConfig](./useRouteconfig/README.md) - Used to convert a page configuration into a useable React-router configuration.
```
matthew-chirgwin: Agree this should be a hook (single responsibility) - but do you/would you expect this to be a re used function (and not say a 'private' hook in the navigation? The summary in this MD mentions the hooks here are expected to be re used. As is, I am unsure if this would be by anything other than the navigation

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  const leaves: Array<Leaf> = [];
+
+  Object.values(input).map((value) => {
+    const page = value as Page;
+    const contexts = page.contexts;
+
+    contexts.map((context) => {
+      // TRANSLATE THIS ONCE 188N IS IN
```
matthew-chirgwin: it is :)

client/Hooks/types.ts
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { FunctionComponent } from 'react';
+
+export type PageProperties = {
```
matthew-chirgwin: I havent gone through the whole PR yet, but out of interest, are these types (expected to be) used in more than one place? With the server code I defined the types seperatley as some were needed in multiple files. Might be simpler if only used in one place to have them defined as a part of the function/object etc (as this file in theory would contain types for N hooks as we add them). Thoughts?

Also, JSDoc?

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
...
+    meta: {},
+  };
+
+  const leaves: Array<Leaf> = [];
+
+  Object.values(input).map((value) => {
+    const page = value as Page;
+    const contexts = page.contexts;
+
+    contexts.map((context) => {
```
matthew-chirgwin: being uber picky, wouldn't a context reduce (ie take array of contexts, reduce them to a single `output` shape) be cleaner? Rather than mapping and changing output by scope?

client/Hooks/useRouteConfig/useRouteConfig.spec.ts
```diff
@@ -0,0 +1,419 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { useRouteConfig } from 'Hooks';
+import { PageConfig, RouterConfig } from 'Hooks/types';
+import HomePageComponent from '../../../test_common/resources/Home.view';
```
matthew-chirgwin: Not an `useRouteConfig.assets.tsx` file? While I see the intent of sharing a mock X page, I rather these be kept with the hook in this case

matthew-chirgwin (changes requested):
Looks good - just have a couple of comments/questions around approach/where the code should sit

.github/workflows/node-pr-jobs-secure.yml
JQrdan: Yep! Hit this when I raised my draft and the client tests failed so the server tests never ran

client/Hooks/README.md
JQrdan: I did have that thought, I just went with a blanket 'Hooks belong in the Hooks folder' approach. Don't disagree that it could easily be part of Nav structure. Although then we would be going against our own approach that 'Bootstrap should only be tested through e2e' :P

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
JQrdan: Ahaha yep! I'll get that with the next set of changes in a future PR

client/Hooks/types.ts
JQrdan: Good point. I was curious as to where a good place to keep these types would be. I quite like them being outside of the source file but I can see that getting massive if it's types for all hooks. Do you think it would be worth having a `Hooks/types` folder that has the types for each file that can be accessed through a barrel file? That way we keep a neat folder structure and can keep the types separately

JQrdan: And yeah, forgot to add JSDoc

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
JQrdan: Will look into doing that 👍

client/Hooks/useRouteConfig/useRouteConfig.spec.ts
JQrdan: I agree with moving these into a single file would be nice. Not so sure about keeping assets next to it though

.github/workflows/node-pr-jobs-secure.yml
nictownsend: https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions - we could do `if success or failure` - means it won't run when a flow is cancelled?

client/Hooks/README.md
nictownsend: I think it's fair to have this as a private hook. Navigation "state" makes sense as context (ability for other components to trigger navigation etc). But this is probably private business logic for the bootstrap to know how to generate page state?

As for unit testing, maybe Bootstrap uses a NavigationLoader that lives elsewhere?

client/Hooks/types.ts
nictownsend: I think it might be worth a type file per hook, then barrel file to export all hooks + types?

client/Hooks/types.ts
```diff
@@ -0,0 +1,69 @@
...
+  mode: string;
+};
+
+export type Context = {
```
nictownsend: https://www.typescriptlang.org/docs/handbook/namespaces.html - namespaces needed, as `Context` seems likely to be a generic name

nictownsend: Also looking at how it's used, I have a horrid feeling it risks confusion with `React.Context`

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { RouterConfig, PageConfig, Page, Leaf } from 'Hooks/types';
+
+const isTopLevelNav = (path) => path.match(/\//g).length === 1;
```
nictownsend: Guessing as this isn't being exported that TS is inferring type based off usage?

nictownsend: Rather than regex, could split on `/` (think it might be more efficient)

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const extractTopLevelPath = (path) => path.match(/\/[^/]*/g)[0];
```
nictownsend: Same here with split

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const generateKey = (name, prefix, properties) =>
+  `${prefix}-${removeWhitespace(removePathVars(name))}${
+    properties && properties.mode ? `.${properties.mode}` : ''
+  }`;
+
+const removeWhitespace = (name) => name.replace(/ /g, '');
```
nictownsend: You need to use `/s` rather than ` ` for true whitespace replace

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const removePathVars = (name) => name.replace(/\${([^}]*)\}/g, '');
```
nictownsend: Is capture group needed?

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
nictownsend: Agree with Matt - also (if we are happy to output es2019):

```
Object.values(input).flatMap(value => value.contexts))
```

will give you an array of contexts that you can then reduce into the output object

client/Hooks/README.md
matthew-chirgwin: Happy to see what the concensus is - I would still vote it should be in with the Navigation myself (unless there is a use case where it is needed eslewhere), and have the tests alongside - no harm in them being there (perhaps no tests are expected, but can be provided if suitable)?

client/Hooks/types.ts
matthew-chirgwin: > Good point. I was curious as to where a good place to keep these types would be. I quite like them being outside of the source file but I can see that getting massive if it's types for all hooks. Do you think it would be worth having a `Hooks/types` folder that has the types for each file that can be accessed through a barrel file? That way we keep a neat folder structure and can keep the types separately
`
If a type is used in multiple files, then it being in a seperate file (`*.assets.ts`) - eg https://github.com/matthew-chirgwin/strimzi-ui/blob/13-integrate/client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.assets.ts  is what I would suggest for now. The users of types will be developers, who likely either want guidance on what X is, or want to type X as something. Thus, I think the types being a constant/exported item from the thing that defines them makes sense. I am not sure all the types being in one file or sub directory would make sense, as it breaks them away from the thing(s) which use them.

client/Hooks/useRouteConfig/useRouteConfig.spec.ts
matthew-chirgwin: When I reviewed, I saw these as test fixtures. We have prevision for such things ins a `*.assets.ts` file - assets is a catch all for constants and things re used across test/storybook. I would class these stubs as such.

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
nictownsend: Or alternatively:
```
Object.values(input).reduce((total, page) => [...total, ...page.contexts])
```

client/Hooks/useRouteConfig/useRouteConfig.spec.ts
JQrdan: Would you expect both types and test stubs to both be part of an `*.assets.ts` file? That seems like something that shouldn't be mixed

test_common/resources/ConsumerGroups.view.tsx
```diff
@@ -0,0 +1,11 @@
+/*
```
nictownsend: I think a generator function might be better instead of duplication - `(componentName) => (): ReactElement => {
  return <div>{componentName}</div>;
}`

client/Hooks/useRouteConfig/useRouteConfig.spec.ts
nictownsend: Agree with asset (given purpose of this) - or a generic generator that can create components in this test (and maybe others - see below comment)

nictownsend (changes requested):

.github/workflows/node-pr-jobs-secure.yml
JQrdan: Didn't realise that always() also runs if cancelled, good call.

client/Hooks/types.ts
JQrdan: @nictownsend Yep! Already making that approach now

client/Hooks/types.ts
JQrdan: Fair point, will rename

client/Hooks/useRouteConfig/useRouteConfig.hook.ts
JQrdan: Yep, the capture group is what ends up getting removed

nictownsend: Unfortunately, no... ;-) (In hindsight, I should have answered my own question and put this as the first comment)

What's being replaced is `${x}` - so `/path/with${some}/${params} -> /path/with/`

The capture group can be referenced in the replacement - so `name.replace(/\${([^}]*)\}/g, '$1')` will change `/path/with${some}/${params} -> /path/withsome/params`.

There's no harm in keeping it, but it's currently unnecessary.

.github/workflows/node-pr-jobs-secure.yml
```diff
@@ -47,13 +47,14 @@ jobs:
       - name: Test client code
         run: npm run test:jest:client -- --forceExit # force exit in case a test does not clean up after failure
       - name: Test server code
+        if: ${{ success() || failure() }}
         run: npm run test:jest:server -- --forceExit # force exit in case a test does not clean up after failure
       - name: Coverage report
         id: coverage
-        if: ${{ always() }}
+        if: ${{ success() || failure() }}
```
nictownsend, line 54: I have a feeling this will fail if `server` step has not completed - there will be no coverage for server

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,95 @@
...
+    properties && properties.mode ? `.${properties.mode}` : ''
+  }`;
+
+const removeWhitespace = (name) => name.replace(/\s/g, '');
+
+const removePathVars = (name) => name.replace(/\${[^}]*\}/g, '');
+
+const contextReducer = (acc, context, page, leaves) => {
+  // TRANSLATE THIS ONCE 188N IS IN
+  const translatedName = context.name;
```
nictownsend, line 23: Coming in future PR?

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -0,0 +1,95 @@
...
+    leaves: [],
+  };
+
+  return acc;
+};
+
+const mergeReduce = (acc, cur) =>
+  mergeWith(acc, cur, (objValue, srcValue) => {
+    if (Array.isArray(objValue)) {
+      return objValue.concat(srcValue);
```
nictownsend, line 65: Do we want to concat or perform a slightly smarter union? (Whether we want duplicates or not)

package.json
```diff
@@ -56,16 +56,16 @@
     "i18next": "^19.8.3",
     "i18next-browser-languagedetector": "^6.0.1",
     "lodash.merge": "^4.6.2",
-    "mini-css-extract-plugin": "^1.2.1",
-    "optimize-css-assets-webpack-plugin": "^5.0.4",
-    "react": "^16.13.1",
+    "lodash.mergewith": "^4.6.2",
```
nictownsend, line 61: I think @pmuir is bringing in another lodash library, wonder if it's worth bringing in full lodash and making more specific requires?

nictownsend (changes requested):

.github/workflows/node-pr-jobs-secure.yml
JQrdan: Looking at the code, yep. It'll fail the step if it can't read the coverage but that would be a bug regardless of this change. Worth raising an issue to handle that a bit better since we should be showing whatever coverage we do have.

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
JQrdan: It sure is

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
JQrdan: It's possible that it would end up with duplicates but you'd have to have a duplicate in the original config in the first place which wouldn't be great

package.json
JQrdan: Possibly! That should come as a dedicated refactor in a future PR, worth another issue.

pmuir (approved):

dlabaj (approved):

client/Bootstrap/Navigation/index.ts
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+export * from './useRouteConfig/useRouteConfig.hook';
```
matthew-chirgwin, line 5: This can change in a latter PR pending discussion - but would we want to expose the hook to anyone? I think this would be private, and thus not available via the barrel file

matthew-chirgwin (approved):
LGTM - one comment re exporting of the `useRouteConfig` hook

scholzj (approved):

client/Bootstrap/Navigation/index.ts
JQrdan: That depends on how we feel about importing stuff even if private? An import from `Navigation` would still be quite nice, even if that component is also in navigation.

matthew-chirgwin: I think it would be confusing. Public pieces only should be in the barrel - as the exposed API.

As an example, the server imports private pieces like this: https://github.com/strimzi/strimzi-ui/blob/5e277d42aa32e734f416347500b4b773df73f260/server/client/router.ts#L7 - it isnt that awkard/describes what is being imported and from where.

Issue #50, opened 2020-10-22
Implement `mockapi` module


matthew-chirgwin:
Issue to cover the implementation of a mock version of `strimzi-admin`.

Implementation idea: Use GitHub submodule to get the `Strimzi-admin` graphql schema, which is used by the mock admin server to generate mock data for development.

PR #94, opened 2020-11-09
feat: implement configuration, context, typing, gql server

### LINKED ISSUE #13: Architecture: Configuration and feature flagging

Investigate and add to the Architecture document from #6 the UI's approach (for both client and server) for feature flagging, and the providing/use of configuration to the UI.

matthew-chirgwin:
- adds configuration types
- adds configuration utilities
- move configuration from server to config
- adds basic context for configuration/feature flags in client
- updates doc
- adds apollo server to express in config and mockapi modules
- starts to flesh out config server module
- starts to flesh out mockapi module
- adds tests
- fixes styling bug in Home component

Contributes to: #13
Contributes to: #50

Signed-off-by: Matthew Chirgwin <chirmatt@uk.ibm.com>


matthew-chirgwin:
There will be a following PR behind this to complete a few outstanding pieces of work:

- look into moving the server types to config
- templating key items into index html
- merging runtime overrides with the defined config
- properly itegrate the config context to call the config endpoint

This however is complete otherwise - enough that others can start to work with it. There is overlap between this issue and #50 - I have added a basic GQL server with test to the mockapi module - but this needs more work (which will be covered in #50)

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <ConfigFeatureFlag.Provider
+      {...others}
+      value={{ ...configInState, ...contextStates }}
+    >
+      {children}
+    </ConfigFeatureFlag.Provider>
+  );
+};
+
+export { ConfigFeatureFlagProvider, ConfigFeatureFlagConsumer };
```
JQrdan, line 42: The React hooks docs says that the full context needs to be used for the useContext hook so can that be exported too. https://reactjs.org/docs/hooks-reference.html#usecontext

matthew-chirgwin: > The React hooks docs says that the full context needs to be used for the useContext hook so can that be exported too. https://reactjs.org/docs/hooks-reference.html#usecontext

Indeed - In the next PR, I will have a `useConfigFeatureFlag` hook (which wraps `useContext`) so the context itself is private, but still accessible

JQrdan: Ok perfect! As long as it's accessible through a hook

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+      const { getByText } = render(
+        <ConfigFeatureFlagProvider>
+          <ConfigFeatureFlagConsumer>
+            {({ config }) => {
+              return <p>{JSON.stringify(config)}</p>;
+            }}
+          </ConfigFeatureFlagConsumer>
+        </ConfigFeatureFlagProvider>
+      );
+      expect(getByText('{"example":{"value":true}}')).toBeInTheDocument();
```
nictownsend, line 25: I'd prefer if we seeded the test with the config, rather than test the inbuilt default? (I see later it's just dummy code but would be good). Unless of course the context should always be in charge of fetching and thus you don't want ability to override it?

Also, I know test docs state describe/it based "tdd" for `client/Contexts` - but wondering whether gherkin actually still applies here?

Or alternatively:

```
const setUp = (renderFn) => 
(<ConfigFeatureFlagProvider>
  <ConfigFeatureFlagConsumer>
     {renderFn}
  </CFFC>
</CFFP>
```

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+      expect(getByText('{"capability":{"enabled":false}}')).toBeInTheDocument();
+    });
+
+    it('returns the expected state values', () => {
+      // check the internals of the context are provided to consumers
+      render(
+        <ConfigFeatureFlagProvider>
+          <ConfigFeatureFlagConsumer>
+            {({ loading, error, isComplete, triggerRefetch }) => {
+              expect(loading).toBe(false);
```
nictownsend, line 47: Have you tried failure cases? Intrigued if expectation throwing error from render function ends the test properly

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,42 @@
...
+  triggerRefetch: () => true,
+};
+
+const ConfigFeatureFlag = createContext({
+  config: {},
+  featureFlags: {},
+  ...contextStates,
+});
+const ConfigFeatureFlagConsumer = ConfigFeatureFlag.Consumer;
+const ConfigFeatureFlagProvider: FunctionComponent = ({
```
nictownsend, line 25: I know this will be changed when you add in the data fetching. But I think it would be good to show an example in here of how the updating will work.

I'm hoping via a mixture of scoping, you will end up calling `triggerRefetch` -> query progress updates the `contextState`  -> updates config/featureFlags -> triggering the `useState` to re-render -> Provider to update

nictownsend: Also, TS typing likely needed here - especially to ensure that the `defaultValue` to the context matches the shape that we later pass to `value` in provider

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import React, { createContext, FunctionComponent, useState } from 'react';
+
+// Temporary config/feature flag definition - to be filled out in later PR with server calls. Add values here and in `strimzi-ui/config` directory
+const config = { example: { value: true } };
+const featureFlags = { capability: { enabled: false } };
+
+// Loading/Error/Trigger Refetch etc
+const contextStates = {
```
nictownsend, line 12: Guessing this will be replaced by a `gql` query hook?

config/configHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import merge from 'lodash.merge';
+import {
+  configurationType,
+  programmaticValue,
+  configurationLiteralValue,
+} from './types';
+
+/** helper function to get the value of an environment variable, or a defined fallback value */
+export const getEnvvarValue: (
```
nictownsend, line 10: What is this sorcery? (might be overkill but https://www.logicbig.com/tutorials/misc/typescript/interface-describing-function.html) defining interface above and using it in the function?

config/configHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+  name: string,
+  fallback?: configurationLiteralValue
+) => () => configurationLiteralValue = (
+  name,
+  fallback = `No value for ENVVAR ${name}`
+) => () => process.env[name] || fallback;
```
nictownsend: Probably fine - but `configurationLiteralValue` is a type union, but process.env will only ever return `string` - which (may) cause some confusion when using it with a fallback that isn't a string.

config/configHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/** helper used to process a `configurationDeclaration` or `featureFlagDeclaration` (D) to a `configurationType` with values being of type (T) for use across the codebase */
+export const processConfig: <D, T>(config: Array<D>) => configurationType<T> = (
+  config
+) => {
+  // merge all provided configs
+  const mergedConfig = config.reduce(
+    (acc, thisConfig) => merge(acc, thisConfig),
```
nictownsend, line 22: Are there any of the following use cases expected:
1. Two configs have the same key and primitive value
2. Two configs have the same key with an array value


If so - merge will override value with the latest (in array case see : https://github.com/lodash/lodash/issues/2872)

config/configHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+  const processedConfiguration = Object.entries(mergedConfig).reduce(
+    (acc, [key, value]) => {
+      const valueType = typeof value;
+      const isLiteralValue =
+        valueType === 'string' ||
+        valueType === 'boolean' ||
+        valueType === 'number';
+      const isConfigurationValue =
+        !isLiteralValue &&
+        (value as programmaticValue<unknown, unknown>).configValue;
```
nictownsend: Won't this return `false` if the `configValue` is falsy? So you probably need undefined check at minimum?

config/configHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+      } else if (isConfigurationValue) {
+        const { configValue, publicConfigValue } = value as programmaticValue<
+          unknown,
+          unknown
+        >;
+        privateValue =
+          typeof configValue === 'function' ? configValue() : configValue;
+        publicValue = publicConfigValue;
+      } else {
+        const { values, publicValues } = processConfig([value]);
```
nictownsend: Being a generic function - what happens if you supply a `D` that has children that don't match `programmaticValue`, `literalValue` or `D`? 

I'm (guessing) typescript will complain - but it means for e.g that if you passed in `{type: ["x"]}` it should either infinite loop or blow up.

nictownsend: ~It also seems to suggest that every returned config value is of type `T` (so I'm guessing you then just have to pass in `T` as `unknown` or a subset.~

Saw later down how you can pass it through. I think the generic does work, but maybe the comment needs a bit more clarity to suggest that every value in `D` needs to be of type `T`

nictownsend: It almost suggests that `D` needs `T` as a parameter somehow too (like `<D, T>(config: Array<D<T>>) => configurationType<T>`)

server/README.md
```diff
@@ -14,3 +14,27 @@ This directory contains all server code for the Strimzi UI - ie code which is re
...
+
+| Configuration                | Required | Default                                                                                                                | Purpose                                                                                                                                                                                |
+| ---------------------------- | -------- | ---------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| authentication.strategy      | No       | `none`                                                                                                                 | What authentication strategy to use to authenticate users. See [the security section](#security) for details of the available options.                                                 |
```
nictownsend, line 25: Overlap with security PR but this needs `authorization.strategy/configuration` too

server/core/app.ts
```diff
@@ -14,9 +14,8 @@ import {
 } from 'logging';
 
 export const returnExpress: (
-  serverName: string,
   getConfig: () => serverConfig
-) => express.Application = (serverName, getConfig) => {
+) => express.Application = (getConfig) => {
```
nictownsend, line 18: Why no server name?

tsconfig.base.json
```diff
@@ -12,7 +12,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "noImplicitAny": false,
-    "outDir": "../../js"
+    "outDir": "./js"
   },
-  "exclude": ["../../utils"]
+  "exclude": ["utils", "build", "**/test_common", "**/*.js"]
```
nictownsend, line 17: As this is being extended by `/client` and `/server` - are the relative paths not needed? (Including `outdir`)

And why are we ignoring `.js` files? I thought we were using `allowJS` to ensure we transpile the compiled JS + any source JS to the right es module support?

server/test_common/testGQLRequests.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+
+/**
+ * Mock requests used to test the gql elements of the server. Query names referenced in feature files, E.g `When I make a 'mockTopicsQuery' gql request to ...`. These can be generated using GraphiQL and it' export feature
+ */
+const mockTopicsQuery = {
+  query: '{\n  topics {\n    name\n    partitions\n    replicas\n  }\n}\n',
```
nictownsend, line 10: Does this need to be a string? Can supertest take an object as input?

nictownsend (changes requested):

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: This will be changing in the next PR, where with the introduction of the apollo pieces, seeding real 'mocked' response data was a requirement for the provider to be tested properly. I also have made some changes to the tests to use more helper code, akin to your `setUp` example in that PR

matthew-chirgwin: https://github.com/matthew-chirgwin/strimzi-ui/tree/13-integrate/client/Contexts/ConfigFeatureFlag

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: The error here is placeholder - from the cases I have seen plumbing in an actual apollo request, no excpetion is thrown - we are instead provided with the error to do with whatever we want. I have tests for this in the next PR

client/Contexts/ConfigFeatureFlag/Context.tsx
matthew-chirgwin: > I know this will be changed when you add in the data fetching. But I think it would be good to show an example in here of how the updating will work.
> 
> I'm hoping via a mixture of scoping, you will end up calling `triggerRefetch` -> query progress updates the `contextState` -> updates config/featureFlags -> triggering the `useState` to re-render -> Provider to update

The flow will be the query will trigger on mount, and when a result (success or otherwise) is returned, a re render will occur for all children.

The refetch I was expecting would only be used in error scenarios, but there is nothing stopping it being used at any time. The result of the refecth will equally trigger a re render.

As for typing, I belive both here and in the next set of changes the typing is provided/correct. The next PR gets a bit more interesting due to the dynamic shape of what comes back, but the shape will follow what is defined in `config/types`

client/Contexts/ConfigFeatureFlag/Context.tsx
matthew-chirgwin: It will indeed - placeholder in case anyone needed it

config/configHelpers.ts
matthew-chirgwin: This is how I have been typing functions! :) A bit involved in this case with the curry, but my general approach is if the typing is not re used, I define it fully alongside/as a part of the function defintion as well. The interface would do the same thing, but in two stages - so if the signiture/type was shared, I would have made it one

config/configHelpers.ts
matthew-chirgwin: Fair point - will update in the next PR

config/configHelpers.ts
matthew-chirgwin: I expect we could have the same key defined in two files, with different values (primitive or otherwise). Eg a static default, and runtime 'real' perhaps. It means the order the configs are provided matter. Think a jsdoc comment would suffice?

config/configHelpers.ts
matthew-chirgwin: Good spot! Happy for this to land in next PR?

config/configHelpers.ts
matthew-chirgwin: > Being a generic function - what happens if you supply a `D` that has children that don't match `programmaticValue`, `literalValue` or `D`?
> 
> I'm (guessing) typescript will complain - but it means for e.g that if you passed in `{type: ["x"]}` it should either infinite loop or blow up.

It will - and with the config we have, the code wont compile without it being the right type (without config changes).

Will look to update typing now - when used I do specify the exact D and T https://github.com/matthew-chirgwin/strimzi-ui/blob/f3feb7ea63fabb60d9ee7e47d11de13e1b25f5f4/config/index.ts#L18 - but agree reading the function in isolation will make this hard to follow

server/README.md
matthew-chirgwin: 👍 - Would this come as a part of the security work? Main motivation with the movement of the config table from architecture to here was to slim down the architecture doc/keep the config with the code it acts on

server/core/app.ts
matthew-chirgwin: It isnt needed! As there is another parameter the linter didnt complain, but this was never used

server/test_common/testGQLRequests.ts
matthew-chirgwin: `These can be generated using GraphiQL and it' export feature`

Suspect using graphql tag it could be a typical query definition

Will fix the it' in the next PR

tsconfig.base.json
matthew-chirgwin: At one stage I ran into an issue with the path config, show did a showConfig tsc run to debug. There I saw all utility, build, test code and linter (symlinks) + jest config being referenced in both the client and server outputs. I don't know how, but it seems when extended, the exclude is witth respect to where the config that defines it is. Hence the traversal didnt work.

This is an additional change - one I can back out. But with it, The config looked as expected for the client and server

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
nictownsend: Apologies, I meant if the `expect` fails in the test, how the test runner handles it

server/README.md
nictownsend: Yep that's fine :)

tsconfig.base.json
nictownsend: I trust you :)

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: It handles it - and doesnt (seemingly) effect other tests:

```
 ● ConfigFeatureFlag › `ConfigFeatureFlagProvider` and `ConfigFeatureFlagConsumer` components › initial state › returns the expected state values on mount

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

      119 |                   rawResponse,
      120 |                 }) => {
    > 121 |                   expect(loading).toBe(false);
          |                                   ^
      122 |                   expect(error).toBe(false);
      123 |                   expect(isComplete).toBe(false);
      124 |                   expect(rawResponse).toEqual({});
```

matthew-chirgwin:
For completeness, I have addressed @nictownsend comments in #99  - which once this is merged, will be be converted to 'full' PR

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+import {
+  ConfigFeatureFlagProvider,
+  ConfigFeatureFlagConsumer,
+} from './Context';
+
+describe('ConfigFeatureFlag context', () => {
+  describe('initial state tests', () => {
+    it('returns the expected configuration values', () => {
+      const { getByText } = render(
+        <ConfigFeatureFlagProvider>
```
JQrdan, line 17: Did you find this was easier than using the provider helper?

JQrdan (approved):
LGTM, nothing to add ontop of Nic's review

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: The `ConfigFeatureFlagProvider` isnt a regular provider - it is a component which makes a query that provides a value - thus I had to drive it as such. As you will see in #99 though, I do use your helper for anything which consumes only :)

scholzj (approved):

matthew-chirgwin:
Any thoughts @pmuir / @dlabaj ?

config/configHelpers.ts
pmuir: I've always use the approach defined in the blog - personally I think it's possibly a bit more readable

pmuir (approved):

Issue #87, opened 2020-11-05
UI Bootstrap UX


matthew-chirgwin:
Issue to cover the discussion and development of UX of the timeframe where the UI bootstraps itself.

From a UX POV, we will need to cover/handle the loading (and error states) of three data requests/result handling. These requests will be for:

- Authorization status (if security enabled)
- Introspection (to discover backend capability)
- Configuration

In a golden path scenario, I imagine this will be a loading spinner (or similar) while the data is fetched/processed. I would expect all 3 calls to be excuted in paralell, and be reasonibly quick to complete.

What I am unsure of is how an error would be surfaced, and if a retry button (for example) would be of use.

As a part of the discussion, the defintion of a set of Gherkin based behaviours should be identified, which will be used as the base for acceptance testing.

This discussion may also identify other UI components required to enable these flows. These should be discussed/designed in their own issues.

chloepoulterdesign:
Hey @matthew-chirgwin is this for any particular user action, i.e. logging in, or is this on any given UI action?

Carbon has very well-written loading guidance available here: https://www.carbondesignsystem.com/patterns/loading-pattern

It talks through when full page spinners should be used, vs skeleton states.
The basic statement is - **the full page load with darkened overlay and spinner is the default.** 
![image](https://user-images.githubusercontent.com/47979346/99983692-167b5500-2da4-11eb-9d83-55205791c206.png)
Then, *if* a page takes a while, use progressive techniques like skeleton states.
![image](https://user-images.githubusercontent.com/47979346/99983784-327ef680-2da4-11eb-9f67-6086f95b9323.png)

matthew-chirgwin:
> Hey @matthew-chirgwin is this for any particular user action, i.e. logging in, or is this on any given UI action?

Sorry, just saw this! This is a one time load, when the user first loads the UI up. There are several things going on under the covers, all of which I would suggest a full page spinner (like the one in the top screenshot) would cover. One thing I am interested to get views on is when some of the under covers work fails -  as this pretty much means the UI will be missing critical info in order to function. I assume an error message/retry buttons would help, but there would not be a lot a user could do potentially.

chloepoulterdesign:
yeah that's an interesting one - I would probably suggest using the 404 HTTP error page - which has a "back to topics" link in it - which I would effectively see behaving as a refresh, at least while topics is the only page.

pmuir:
[...]

PR #99, opened 2020-11-12
Complete Apollo client integration, add FeatureFlag component

### LINKED ISSUE #13: Architecture: Configuration and feature flagging

matthew-chirgwin:
- Fleshes out the config context to retrieve from the server, and stores the result in context
- Inserts bootstrap config and provides it via context
- Adds feature flagging components
- Tests for all areas

Known issues:

~400/404 errors will appear in the browser console as the UI attempts to connect to the `/api` endpoint. I have a fix for this, which will follow in another PR (pending #85)~ - in the interest of time, I have added a small change to make all GQL requests via HTTP - this stops errors from being thrown/enables UI development transpartent to the transport being used. I will follow up with a WS change in another PR


client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,382 @@
...
+                  isComplete,
+                  triggerRefetch,
+                  rawResponse,
+                }) => {
+                  expect(loading).toBe(true);
+                  expect(error).toBe(false);
+                  expect(isComplete).toBe(false);
+                  expect(rawResponse).toEqual({});
+                  expect(triggerRefetch).toEqual(expect.any(Function));
+                  return <p>{JSON.stringify(isComplete)}</p>;
```
JQrdan: Is there a world where these expects don't run so the test passes anyway? Is it worth passing the done parameter in so we know for sure that the test has completed because these asserts have happened?

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,60 @@
...
+} from './ConfigFeatureFlag.assets';
+
+const ConfigFeatureFlag = createContext(defaultConfigFeatureFlagValue);
+/** useConfigFeatureFlag - hook which returns the current value of the ConfigFeatureFlag context */
+const useConfigFeatureFlag: () => ConfigFeatureFlagType = () =>
+  useContext(ConfigFeatureFlag);
+/** ConfigFeatureFlagConsumer - consumer component of the ConfigFeatureFlag context */
+const ConfigFeatureFlagConsumer = ConfigFeatureFlag.Consumer;
+/** Test provider. Only to be used in tests */
+const _ConfigFeatureFlag_TestProvider = ConfigFeatureFlag.Provider;
```
JQrdan: Code specific for tests in the real implementation? 😬

client/Contexts/ConfigFeatureFlag/README.md
```diff
@@ -0,0 +1,31 @@
+# ConfigFeatureFlag
+
+This context will perform and store the result of a GraphQL query against the `/config` endpoint presented by the UI server, returning all configuration and feature flag state. This will allow the UI client discover and refresh itself at runtime. The context will provide hooks and consumers to direct access both the feature flag and config values.
```
JQrdan:
```suggestion
This context will perform and store the result of a GraphQL query against the `/config` endpoint presented by the UI server, returning all configuration and feature flag state. This will allow the UI client to discover and refresh itself at runtime. The context will provide hooks and consumers to directly access both the feature flag and config values.
```

client/Contexts/ConfigFeatureFlag/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Usage
+
+This context provides 4 publicly exported items for use. They are the context producer, consumer and hook, as well as a `FeatureFlag` component to provide a simple declarative method to enable or disable specific UI sections.
+
+### `ConfigFeatureFlagProvider`, `ConfigFeatureFlagConsumer` and `useConfigFeatureFlag`
+
+- `ConfigFeatureFlagProvider` is a component which makes the request to the `/config` endpoint, and manages the result. It should only be placed in the `index.ts` in `Bootstrap` as it is intended to be a single use component.
+- `ConfigFeatureFlagConsumer` is a standard Context consumer, providing the `value` of the context at render time.
+- `useConfigFeatureFlag` is a hook, which provides the `value` of the context at render time.
```
JQrdan: it would be nice to provide an example of how to use the `useconfigFeatureFlag` hook here as well as the component

client/Panels/Home/Home.tsx
```diff
@@ -4,12 +4,19 @@
...
+  const { client, isComplete } = useConfigFeatureFlag();
+
   return (
-    <div>
+    <div className='home'>
       <img src={image} alt='Strimzi logo' />
       Welcome to the Strimzi UI
+      <FeatureFlag flag={'client.Home.showVersion'}>
+        {isComplete && `Version: ${client.version}`}
+      </FeatureFlag>
```
JQrdan: I'm not so sure about the mixing of a hook and a component to achieve this. Is there a way where the hook itself can give you enough information so that a developer could just use that result to decide what to render?

JQrdan (changes requested):

client/Panels/Home/Home.tsx
JQrdan: Sorry I've just noticed these are two separate ways of using this!

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,382 @@
...
+      const { triggerRefetch, ...others } = useConfigFeatureFlag();
+      return (
+        <p>
+          {JSON.stringify({
+            ...others,
+            triggerRefetch: typeof triggerRefetch === 'function',
+          })}
+        </p>
+      );
+    };
```
JQrdan: Is there a reason you're doing it this way instead of using https://www.npmjs.com/package/@testing-library/react-hooks?

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  apolloQueryResponseType,
+  ConfigFeatureFlagType,
+  defaultClientConfig,
+  defaultConfigFeatureFlagValue,
+} from './ConfigFeatureFlag.assets';
+
+const ConfigFeatureFlag = createContext(defaultConfigFeatureFlagValue);
+/** useConfigFeatureFlag - hook which returns the current value of the ConfigFeatureFlag context */
+const useConfigFeatureFlag: () => ConfigFeatureFlagType = () =>
```
JQrdan: Opinions on this being exposed through the hooks folder instead?

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: Did not realise this was a thing! Will update to use it

client/Contexts/ConfigFeatureFlag/Context.tsx
matthew-chirgwin: Fair point - think this makes sense, and also means I can get rid of the 'private' provider as I will now need to export the context as a whole (this can use the provider from that). Will update

client/Panels/Home/Home.tsx
matthew-chirgwin: Yeah - a user could use the hook, or the component, whatever makes sense. I can see cases where both would be helpful

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: If the expect fails, the error bubbles up and fail the test - see https://github.com/strimzi/strimzi-ui/pull/94#discussion_r521996835

utils/test/context/context.util.tsx
```diff
@@ -4,46 +4,75 @@
...
   render(ui, {
     wrapper: ({ children }) => {
       return contextWrapper(providers, children);
     },
     ...options,
   });
 
-export { renderWithContextProviders };
+/** withConfigFeatureFlagContext renders the given `ui` JSX with a ConfigFeatureFlag provider seeded with default values. Use this if your components indirectly use/require a value from the ConfigFeatureFlag context, else use `withCustomConfigFeatureFlagContext` so the value from ConfigFeatureFlag can be controlled  */
+const withConfigFeatureFlagContext: (
```
JQrdan: Given the naming of these functions, I'd assume that they would be used like:

```
render(withConfigFeatureFlagContext(...))
```

But the render function isn't needed. Could these be modified so they work in that way or renamed to `renderWithCustomConfigFeatureFlagContext` so that the names are consistent with the behaviour

matthew-chirgwin: Agreed - will be updated in the next commit

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+import {
+  ConfigFeatureFlagProvider,
+  ConfigFeatureFlagConsumer,
+} from './Context';
+import { defaultClientConfig } from './ConfigFeatureFlag.assets';
+
+import { ConfigFeatureFlagType } from './ConfigFeatureFlag.types';
+
+describe('ConfigFeatureFlag', () => {
+  const mockClientResponse = {
```
nictownsend: Move to `assets`?

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+      result: {
+        data: {
+          client: mockClientResponse,
+          featureFlags: mockFeatureFlagResponse,
+        },
+      },
+    },
+  ];
+
+  const noOpResponse = [
```
nictownsend:
```suggestion
  const noOpQueryMock = [
```

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+      request,
+      result: {
+        data: {},
+      },
+    },
+  ];
+
+  describe('`ConfigFeatureFlagProvider` and `ConfigFeatureFlagConsumer` components', () => {
+    describe('initial state', () => {
+      it('returns the expected client configuration values', () => {
```
nictownsend:
```suggestion
      it('returns the default client configuration values', () => {
```

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+        const expectedValues = defaultClientConfig.client;
+        const { getByText } = render(
+          withApolloProviderReturning(
```
nictownsend:
```suggestion
          setup(gqlMock, renderFn)
```
Where setup can create the `withApolloProvider` duplication

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
nictownsend: Can you extract this into an "expectations" render function so it's not nested inside the component here

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+                }}
+              </ConfigFeatureFlagConsumer>
+            </ConfigFeatureFlagProvider>
+          )
+        );
+      });
+    });
+    describe('response handling logic', () => {
+      // define common expected states, namespaces so RTL getByText can identify them
+      const namespaceValue: (
```
nictownsend: Common `stringifyForRender` or similar?

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+            noOpResponse,
+            <ConfigFeatureFlagProvider>
+              <ConfigFeatureFlagConsumer>
+                {testRenderFunc}
+              </ConfigFeatureFlagConsumer>
+            </ConfigFeatureFlagProvider>
+          )
+        );
+        act(() => {
+          expect(getByText(JSON.stringify(expectedClient))).toBeInTheDocument();
```
nictownsend, line 224: https://testing-library.com/docs/dom-testing-library/api-async - `waitFor` of any use here?

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
```diff
@@ -0,0 +1,347 @@
...
+        const expectedClient = emptyClientResponse;
+        const expectedFeatureFlags = emptyFeatureFlagResponse;
+        const expectedStatusLoading = initialLoadingStatus;
+        const expectedStatusError = errorStatus;
+        const { getByText } = render(
+          withApolloProviderReturning(
+            errorResponse,
+            <ConfigFeatureFlagProvider>
+              <ConfigFeatureFlagConsumer>
+                {testRenderFunc}
```
nictownsend: If we control render function, is the `expect` inside render better than having to stringify and use `getByText`?

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.types.ts
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { ApolloError, QueryResult } from '@apollo/client';
+import { exposedClientType, exposedFeatureFlagsType } from 'ui-config';
+
+/** the response (`data`) value from the graphql config query (client/Queries/Config/index.ts) */
+export type apolloQueryResponseType = {
```
nictownsend: General question - should we treat types like classes and have CamelCase as a naming strategy?

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+  });
+  const { client, featureFlags }: apolloQueryResponseType =
+    data || defaultClientConfig;
+  const value: ConfigFeatureFlagType = {
+    client,
+    featureFlags,
+    loading,
+    error: error ? true : false,
+    isComplete: !(loading || error) && !loading && !error,
```
nictownsend:
```suggestion
    isSuccess: !(loading || error),
```

Might be going crazy but can't see why the second case is needed?

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -0,0 +1,57 @@
...
+    triggerRefetch: () => {
```
nictownsend, line 79: Why is this function nested and not just an alias for `refetch`?

client/Contexts/ConfigFeatureFlag/FeatureFlag.view.tsx
```diff
@@ -0,0 +1,28 @@
...
+  children,
+  flag,
+  ...others
+}) => {
+  const { featureFlags } = useConfigFeatureFlag();
+
+  const featureFlagValue = get(featureFlags, flag, false);
+
+  return (
+    <div {...others} data-ff={flag}>
```
nictownsend, line 24: Is it more or less efficient to return null without the `<div>` when the flag is disabled?

matthew-chirgwin:
This PR now rebased on master - will rebase again once #85 goes in as will need some of it's WS work to integrate Apollo fully

client/Panels/Home/Home.steps.tsx
```diff
@@ -15,12 +16,25 @@ Given('a Home component', () => {
 });
 
 When('it is rendered', () => {
-  renderResult = render(component);
+  renderResult = renderWithCustomConfigFeatureFlagContext(
```
nictownsend, line 32: Can we have a separate test with flags disabled?

utils/test/context/context.util.tsx
```diff
@@ -36,4 +42,37 @@ const renderWithContextProviders: <T>(
...
+) => RenderResult = (ui, options = {}) =>
+  renderWithContextProviders(ui, options, [
+    {
+      provider: ConfigFeatureFlag.Provider,
+      value: defaultConfigFeatureFlagValue,
+    },
+  ]);
+
+/** renderWithCustomConfigFeatureFlagContext renders the given `ui` JSX with a ConfigFeatureFlag provider provided via `configFeatureFlagValue`. */
+const renderWithCustomConfigFeatureFlagContext: (
```
nictownsend, line 58: Can these be composed? I.e can I perform `renderWithConfigFeature(renderWithCustomConfig)` - or a builder esque pattern that allows you to do `withX().withY().withZ().render()`?

client/Panels/Home/Home.steps.tsx
nictownsend: https://github.com/kentcdodds/react-test-context-provider?

nictownsend: Also raises an interesting question - should components with context be tested without a context (to check default behaviour)?

And I'm guessing it means that we have to have a mixture of `prop` and `context` test cases?

client/Panels/Home/Home.tsx
```diff
@@ -3,14 +3,22 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 import React, { FunctionComponent } from 'react';
+import get from 'lodash.get';
 import image from 'Images/logo.png';
 import './style.scss';
+import { useConfigFeatureFlag } from 'Hooks';
 
 const Home: FunctionComponent = ({ children }) => {
+  const { client, featureFlags, isComplete } = useConfigFeatureFlag();
```
nictownsend, line 12: Is the design that any component wanting to use flags will need loading indicators?

client/Queries/Config/index.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import gql from 'graphql-tag';
+
+export const GET_CONFIG = gql`
+  query {
+    client {
```
nictownsend, line 9: Semantically, `about` doesn't feel like "config" - is this more designed to get "runtime" info?

client/Queries/Config/index.ts
```diff
@@ -0,0 +1,25 @@
...
+      about {
+        version
+      }
+    }
+    featureFlags {
+      client {
+        Home {
```
nictownsend, line 16: Should this be a result of introspection? Or alternatively, in gql something like

`client: [featureFlag]`
`type featureFlag = {name:value}`

Perhaps with nesting (so we can define a shape)?

config/featureflags.ts
```diff
@@ -2,10 +2,19 @@
...
+import { Config } from './config.types';
 
 /**
  * feature flag configuration - used to enable/disable capabilities across the UI/server. The values can be static or dynamic, but the value must always be boolean
  */
 
-export const featureFlags: Config<boolean> = {};
+export const featureFlags: Config<boolean> = {
+  client: {
+    Home: {
```
nictownsend, line 13: Same concern over having to define these (they're more like `ConfigValue<boolean>`)

nictownsend: It feels like the UI should not have to know about the flag structure to query for it - and same with server, it parses some feature flag config and then has a key:value structure rather than having to define each feature flag.

package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.1",
   "description": "Strimzi UI provides a way for managing Strimzi and Kafka clusters (+ other components) deployed by it using a graphical user interface.",
   "scripts": {
-    "clean": "rm -rf dist/ && rm -rf js/ && rm -rf .out/ && rm -rf generated/ && rm -rf coverage/",
+    "clean": "rm -rf dist/ && rm -rf js/ && rm -rf .out/ && rm -rf generated/ && rm -rf coverage/ && npx jest --clear-cache",
```
nictownsend, line 7: Assumes `npx` is installed. You can instead have a `npm run clean-cache` that calls `jest --clear-cache`

server/client/router.ts
```diff
@@ -31,6 +32,13 @@ export const ClientModule: UIServerModule = {
     // add the auth middleware to all non public files
     protectedFiles.forEach((file) => routerForModule.get(`${file}`, authFn));
 
+    // return index.html, with configuration templated in
+    hasIndexFile &&
+      routerForModule.get(
+        '/index.html',
+        renderIndexHtmlWithBootstrapConfig(indexFile)
```
nictownsend: `renderTemplate` maybe is less verbose?

server/config/controller.ts
```diff
@@ -3,48 +3,65 @@
...
-    ({ schemaTypes, resolvers }, { type, typeValue }) => {
-      const { value } = jsonToSchema({
-        baseType: type,
-        jsonInput: JSON.stringify(typeValue),
-      });
-      const allResolvers = { ...resolvers.Query, [type]: () => typeValue };
+export const returnSchemaAndResolversGivenConfig: (
+  config: serverConfigType
+) => configSchemaResolverType = (config) => {
+  const { featureFlags: featureFlagOverrides } = config;
```
nictownsend:
```suggestion
  const { featureFlags: featureFlagOverrides, client: {configOverrides} } = config;
```

server/config/controller.ts
```diff
@@ -3,48 +3,65 @@
...
+  const { configOverrides } = config.client;
 
-      return {
-        schemaTypes: schemaTypes.concat([value]),
-        resolvers: {
-          Query: {
-            ...allResolvers,
+  const configToHost = {
+    featureFlags: merge(
+      merge({}, featureFlags.publicValues),
```
nictownsend: Why the extra `merge`? `merge` is variadic so you can do `merge({}, x, y, z)`

server/config/router.ts
```diff
@@ -7,17 +7,19 @@ import { UIServerModule } from 'types';
...
 export const ConfigModule: UIServerModule = {
   moduleName,
-  addModule: (logger, authFn) => {
+  addModule: (logger, authFn, config) => {
     const { exit } = logger.entry('addModule');
     const routerForModule = express.Router();
 
-    const server = new ApolloServer({ typeDefs: schema, resolvers });
+    const server = new ApolloServer({
+      ...returnSchemaAndResolversGivenConfig(config),
```
nictownsend:
```suggestion
      ...apolloConfig(config),
```

nictownsend (changes requested):

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: I am going to move this to assets, and also add a helper to avoid duplicating response structure

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.types.ts
matthew-chirgwin: I don't have a preference to which, but they should be consistent. Will update `apolloQueryResponseType` to be `ApolloQueryResponseType`

client/Contexts/ConfigFeatureFlag/Context.tsx
matthew-chirgwin: Waiting for logging line to be added. Otherwise, yes it could be

client/Contexts/ConfigFeatureFlag/Context.tsx
matthew-chirgwin: I think you are right! Will remove the first case as feel the second is more readable. As for the name, would prefer (semantically) to keep the `isComplete` notion for readability of consumers of the context

client/Contexts/ConfigFeatureFlag/FeatureFlag.view.tsx
matthew-chirgwin: From a rendered elements POV yes - I was speculating on the helpfulness is debug/test to confirm the container had no content, ie the flag was off

client/Panels/Home/Home.steps.tsx
matthew-chirgwin: Will add a test with flags off now.

Interesting question re testing without context. Contexts will always return a default value - I would tend to have a test which perhaps (via `renderWithConfigFeatureFlagContext`) checks behaviour of this component using the context with default values.

client/Panels/Home/Home.tsx
matthew-chirgwin: As they are retieved from the server, yes, they may need to. This, along with introspection and authz (if needed) would be covered (I suggest) under a single loading state 'on mount'/bootstrap , covered by UX issue #87

client/Queries/Config/index.ts
matthew-chirgwin: I wanted to provide an example of configuration (settings or values) being used in this PR. All of these settings or values would be returned at runtime - I don't mind the name changing if you can think of something better! :)

client/Queries/Config/index.ts
matthew-chirgwin: Introspection is of the Strimzi-admin capability, and if something is enabled (or not) - this is UI capability. So similar, but different.

Re the typing/structure - as the schema is defined dynamically from the config, this is what we get OOTB. I would suggest this is ok as is, but could be changed in future if needed.

package.json
matthew-chirgwin: Not an issue as it is included with the minimum version of npm we reference as a requriement

utils/test/context/context.util.tsx
matthew-chirgwin: Agree it would be nice to have something more fluent, but I was building upon what was already there. Follow on issue?

server/config/controller.ts
matthew-chirgwin: Aha! Much nicer! Will update (along with others)

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: Aha! Yep, looks like a good fit, will update

client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx
matthew-chirgwin: Potentially - I went with this style given the setup/do/assert pattern where possible

dlabaj (approved):

client/Contexts/ConfigFeatureFlag/Context.tsx
```diff
@@ -2,41 +2,95 @@
...
+        : defaultConfigFeatureFlagValue.bootstrapConfig;
+  } else {
+    bootstrapConfig = defaultConfigFeatureFlagValue.bootstrapConfig;
+  }
+
+  const { client, featureFlags }: ApolloQueryResponseType =
+    data || defaultClientConfig;
+
+  // check to see/merge any feature flag state with any defined in browser
+  const sanatisedParams = sanatiseUrlParams(getLocation().search);
```
ajborley: Typo: `sanitised`
```suggestion
  const sanitisedParams = sanitiseUrlParams(getLocation().search);
```

matthew-chirgwin: Thanks Andy - will fix (and rename all files etc) in a new commit

JQrdan (approved):

matthew-chirgwin:
Thank you for reviews/comments @JQrdan @dlabaj  @nictownsend @ajborley - I have just pushed a commit to change the transport to HTTP for now, while I work on the WS support. Can I please ask for another review, with a view to merge this asap as long as there are no objections?

JQrdan (approved):
HTTP changes also look good to me

scholzj (approved):

client/Queries/Config/index.ts
pmuir: I think the config name is fine and fairly standard.

client/Queries/Config/index.ts
pmuir: Agreed, feature flags are orthogonal to introspection.

config/featureflags.ts
pmuir: I think it's better have an explicit schema for the feature flags (which I think this is providing). A freeform map is quite hard to manage in practice, as you don't know the total set of flags you have in the app.

pmuir (approved):
LGTM - note I only did a review of the semantics of the feature flags and apollo client integration, I didn't attempt to a code review.

Issue #20, opened 2020-09-18
A11y testing and support


matthew-chirgwin:
The Strimzi-ui should be fully accessible (a11y) from day 0. This issue is to investigate what tooling we can use to catch a11y issues early/make sure the UI is accessible to any user of it.

As per the proposal (https://github.com/strimzi/proposals/pull/6), we will target the superset of US Section 508 and WCAG21AA rulesets.

nictownsend:
https://www.npmjs.com/package/cypress-accessibility-checker (if not already aware)

PR #102, opened 2020-11-12
DRAFT: Patternfly a11y work.

dlabaj:
A11y work demoing PatternFly a11y tooling.  Fixes issue #20

Signed-off-by: Donald Labaj dlabaj@redhat.com


nictownsend:
Questions:

1. If a page has multiple loading stages, can this be configured to handle them? And if so, does it mean it would need to be configured at a page by page level?
2. How does crawling work? Is it just scraping for `href` and then queues them to try?
3. If a page has multiple dynamic pieces, does this support nested scripts for puppeteer to know how to navigate it?

I like the concept of this crawling - the downside with Cypress is that it requires you to remember to test a page. However, Cypress has advantage of letting you explore the page and check a11y at different times.

Perhaps we could have an approach that merges the two, but I have a feeling the answer to question 1 will mean the crawling won't function well enough to justify using it alongside cypress.

dlabaj:
@nictownsend 

**If a page has multiple loading stages, can this be configured to handle them? And if so, does it mean it would need to be configured at a page by page level?**
We are waiting 3s after the last network request for React to load + render

**How does crawling work? Is it just scraping for href and then queues them to try?**
Yes it looks for anchor tags with relative hrefs.

**If a page has multiple dynamic pieces, does this support nested scripts for puppeteer to know how to navigate it?**
No

nictownsend:
Believe decision was to close this in favour of the cypress work. @dlabaj please close if you agree

dlabaj:
Closing this PR in favor of cypress.

PR #103, opened 2020-11-12
Added Cypress Axe.

dlabaj:
Fixes issue #20 using cypress axe.

Signed-off-by: Donald Labaj dlabaj@redhat.com


e2e/features/step_definitions/coreUX.stepdef.ts
```diff
@@ -6,8 +6,10 @@ import { Given, Then } from 'cypress-cucumber-preprocessor/steps';
 
 Given('I am on the strimzi-ui homepage', () => {
   cy.visit('localhost:3000/index.html');
+  cy.injectAxe();
 });
 
 Then(`the welcome message appears`, () => {
   cy.get('#root').contains(`Welcome to the Strimzi UI`);
+  cy.checkA11y();
```
nictownsend: https://www.npmjs.com/package/cypress-axe#using-the-violationcallback-argument - I'd like if we had support for this to log out to console.

Otherwise, when test fails - while we can watch the video that GHA uploads as an artifact, it would be good to have failure in the workflow log.

nictownsend: Also - Carbon is not always 100% accessible - how do you flag particular rules as "ignore for this element"?

nictownsend: Finally - we will need to discuss "what rules should we run" and "what impact level fails". Can you propose a default set?

e2e/features/step_definitions/coreUX.stepdef.ts
```diff
@@ -5,9 +5,10 @@
 import { Given, Then } from 'cypress-cucumber-preprocessor/steps';
 
 Given('I am on the strimzi-ui homepage', () => {
-  cy.visit('localhost:3000/index.html');
+  cy.visit('localhost:8080/index.html');
```
nictownsend: Why the port change? When your run `npm run e2e` it should be starting up express on `3000`

e2e/support/index.ts
```diff
@@ -2,3 +2,57 @@
...
+      } ${violations.length === 1 ? 'was' : 'were'} detected ${
+        target ? `for ${target}` : ''
+      }`
+    );
+    cy.log('logTable', violationData);
+  }
+);
+
+Cypress.Commands.add('testA11y', (target: string) => {
+  cy.injectAxe();
```
nictownsend, line 47: Can the `inject` and `configure` be run multiple times inside a test (expect there will be multiple calls to `testA11y` - or does it have to be ran only once per test?

e2e/support/index.ts
```diff
@@ -2,3 +2,57 @@
...
+  cy.configureAxe({
+    rules: [
+      { id: 'color-contrast', enabled: false }, // seem to be somewhat inaccurate and has difficulty always picking up the correct colors, tons of open issues for it on axe-core
+      { id: 'focusable-content', enabled: false }, // recently updated and need to give the PF team time to fix issues before enabling
```
nictownsend, line 51: Are the issues being fixed in Axe by PF, or are rules disabled until new version of PF is released?

package.json
```diff
@@ -63,9 +63,8 @@
     "pino-filter": "^1.0.0",
     "pino-http": "^5.3.0",
     "pino-pretty": "^4.3.0",
-    "sass": "^1.29.0",
-    "react-i18next": "^11.7.3",
     "sass": "^1.26.10",
```
nictownsend: When you merge `package.json` please ensure only one `sass` entry and version `1.29.0`

nictownsend (changes requested):
Principle looks good; I'm still unclear as to how you can disable individual rules/elements per `checkA11y` call though. Can you investigate please

e2e/features/step_definitions/coreUX.stepdef.ts
dlabaj: Was a mistake I will fix it

package.json
dlabaj: Thought I fixed that thanks for the catch.

nictownsend:
@dlabaj few outstanding questions that are blocking this being merged (post-rebase)

e2e/support/index.ts
dlabaj: Currently they are being fixed by the patternfly team

e2e/support/index.ts
dlabaj: It only has to be run once.

e2e/features/step_definitions/coreUX.stepdef.ts
dlabaj: We can disable rules that the frameworks don't support.  On patternfly we try to not disable rules.

e2e/support/index.ts
nictownsend: `cy.testally` would be called multiple times - so should `cy.injextAxe()` and `configureAxe()` instead be in a before all?

nictownsend (changes requested):

e2e/support/index.ts
dlabaj: yeah that makes sense I'll modify it.

PR #116, opened 2020-11-16
feat: add express session

### LINKED ISSUE #107: Session support to store user credentials
Express session support - so that a multiple requests can be considered a user session

Use cookie support for session; `cookie.sid` contains reference to session, express session deserializes from cookie id

Memory store for session initially

nictownsend:
- in memory session
- cookie keyed by server name

Contributes to: strimzi/strimzi-ui#107


server/core/app.ts
```diff
@@ -26,6 +27,16 @@ export const returnExpress: (
   // add pino-http middleware
   app.use(generateHttpLogger());
 
+  //Add session middleware
+  const sessionOpts: session.SessionOptions = {
+    secret: 'CHANGEME',
+    name: serverName,
+    cookie: {
+      maxAge: 1000 * 3600 * 24 * 30,
```
JQrdan: Could you add a comment to say how long this is

server/core/app.ts
```diff
@@ -26,6 +27,16 @@ export const returnExpress: (
   // add pino-http middleware
   app.use(generateHttpLogger());
 
+  //Add session middleware
+  const sessionOpts: session.SessionOptions = {
```
JQrdan: Would be neater to import SessionOptions independently

JQrdan (changes requested):
Just a couple of minor things but otherwise LGTM

pmuir (approved):

server/core/app.ts
```diff
@@ -26,6 +27,16 @@ export const returnExpress: (
   // add pino-http middleware
   app.use(generateHttpLogger());
 
+  //Add session middleware
+  const sessionOpts: SessionOptions = {
```
matthew-chirgwin, line 32: I assume (in later issues) this will be enabled/added only when needed?

server/core/core.steps.ts
```diff
@@ -37,90 +37,65 @@ When(
...
+    };
 
-        Object.entries(headerExpectedValue).forEach(([key, value]) =>
-          expect(headers[key]).toEqual(value)
-        );
-      },
-      (err) => {
-        throw err;
-      }
+    return Object.entries(headerExpectedValue).reduce(
```
matthew-chirgwin: This is a lot more subtle/less readable than the prior version IMO. Can a comment be added to explain what this is doing/how the `chainedRequest` is being invoked?

server/test_common/commonServerSteps.ts
```diff
@@ -67,7 +69,22 @@ And(
   stepWhichUpdatesWorld((world) => {
     return {
       ...world,
-      server: request(returnExpress('test-server', () => world.configuration)),
+      server: request(
+        returnExpress(
+          world.serverName || 'test-server',
```
matthew-chirgwin: WIth the removal of the name argument this will need to be updated

matthew-chirgwin (changes requested):
LGTM - just one test needs a comment for me to (when we return to this in future) we can see how the assertions are being chained

server/test_common/commonServerSteps.ts
nictownsend: Just spotted it :'(

server/core/core.steps.ts
nictownsend: I'll revert - if it needs a comment then the efficiency isn't worth the readable impact

scholzj (approved):

dlabaj (approved):

config/static.ts
```diff
@@ -36,6 +36,9 @@ const server: Config<Literal> = {
         port: 9080,
         transport: {},
       },
+      session: {
```
matthew-chirgwin, line 39: no issue with the being here - can the typing and doc be updated to reflect? https://github.com/matthew-chirgwin/strimzi-ui/tree/master/server#configuration-options / https://github.com/matthew-chirgwin/strimzi-ui/blob/master/server/types.ts

matthew-chirgwin (commented):
LGTM - just one comment re typing/doc

server/core/app.ts
nictownsend: I think it's fair to always have a session as we might want to store information for the lifetime of a browser session. In particular, for logging - @ajborley, would it be useful to store a "client id" in the session?

config/static.ts
nictownsend: Adding

server/types.ts
```diff
@@ -79,6 +83,7 @@ export type serverConfig = {
   hostname: string;
   /** The port number to use/accept traffic on */
   port: number;
+  session: sessionConfigType;
```
matthew-chirgwin, line 88: JSDoc for usage/context/purpose?

matthew-chirgwin (approved):

PR #118, opened 2020-11-16
chore: refactor server tests

### LINKED ISSUE #117: Chore: Simplify server tests
https://github.com/visionmedia/supertest#api

SuperTest api supports series of basic expectations on the request. It also returns a promise.

As a result - server tests do not need to use `async/await` unless they need to perform expectations after the supertest response checks.

nictownsend:
- simplify tests by using supertest expectations
- remove unnecessary async functions

Contributes to: strimzi/strimzi-ui#117

### DIFF
--- a/server/api/api.steps.ts
+++ b/server/api/api.steps.ts
@@ -90,7 +90,6 @@ Then(
 
     const expectFailure = context.responseWillError ? true : false;
 
-    // check mock call counts
     await request.expect((res) => {
       // confirm the API module handles the response
       if (expectFailure) {
--- a/server/api/api.steps.ts
+++ b/server/api/api.steps.ts
@@ -91,38 +91,34 @@ Then(
     const expectFailure = context.responseWillError ? true : false;
 
     // check mock call counts
-    await request.then(
-      (res) => {
-        // check all mock calls for expected values
-        expect(createProxyServer).toHaveBeenCalledTimes(1);
-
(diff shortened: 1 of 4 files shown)


server/api/api.steps.ts
```diff
@@ -91,38 +91,34 @@ Then(
     const expectFailure = context.responseWillError ? true : false;
 
     // check mock call counts
```
pmuir: This comment probably needs to go.

pmuir: Or moved donw?

pmuir (approved):

matthew-chirgwin (approved):
LGTM - no objection to the change - it achieves the same assertions as before. I am curious to what benefit using `request.expect` brings (other than the removal of async/await and using it to deal with the async side of things)

nictownsend:
> LGTM - no objection to the change - it achieves the same assertions as before. I am curious to what benefit using `request.expect` brings (other than the removal of async/await and using it to deal with the async side of things)

My thoughts were:
1. Async/await worked if we needed to wait for the response, but as `expect(res => {})` handled all those cases, it seemed more confusing to `await` the request promise but do nothing with it, than return it for jest to handle
2. Removing the `then` got rid of the "catch and rethrow" behaviour
3. The `request.expect` function are mostly shorthand, but it simplified a lot of cases that just cared about headers or status codes

scholzj (approved):

PR #126, opened 2020-11-20
feat: add support for translation and feature flags in navigation

JQrdan:
Contributes to: #68


__mocks__/react-i18next.tsx
```diff
@@ -2,7 +2,7 @@
  * Copyright Strimzi authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
-import { translate, translateWithFormatting } from 'utils/test/i18n';
+import { translate, translateWithFormatting } from '../utils/test';
```
matthew-chirgwin, line 5: out of interest, can the mocks not make use of the path resolution?

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -3,7 +3,11 @@
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
 import { RouterConfig, PageConfig, Page, Leaf } from './useRouteConfig.types';
+import { useTranslation } from 'react-i18next';
 import mergeWith from 'lodash.mergewith';
+import { useConfigFeatureFlag } from 'Hooks';
+import { Error404 } from 'Panels/Error404';
```
matthew-chirgwin: `import { Error404 } from 'Panels';`?

__mocks__/react-i18next.tsx
JQrdan: They can, but an IDE will complain that the path can't be resolved because it would only work when jest subs the file into the right place

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
JQrdan: Auto import ;) will fix

__mocks__/react-i18next.tsx
JQrdan: But this is only the case for mocking anything from `node_modules`. Mocks for our own stuff will be in a path where tsconfig.json exists

nictownsend: Top level tsconfig now worth it?

nictownsend: Or alternatively, leave as is but accept the IDE problems?

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -18,44 +22,61 @@ const removeWhitespace = (name) => name.replace(/\s/g, '');
 
 const removePathVars = (name) => name.replace(/\${[^}]*\}/g, '');
 
-const contextReducer = (acc, context, page, leaves) => {
-  // TRANSLATE THIS ONCE 188N IS IN
-  const translatedName = context.name;
+const checkFeatureFlag = (page, context, featureFlags) => {
+  if (get(featureFlags.client, context.feature_flag)) {
```
nictownsend: Can this return `boolean` instead? (See later comment for use)

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -18,44 +22,61 @@ const removeWhitespace = (name) => name.replace(/\s/g, '');
...
+  context,
+  page,
+  leaves,
+  featureFlags,
+  translate
+) => {
+  const translatedName = translate(context.name);
   const linkKey = generateKey(translatedName, 'link', context.properties);
   const routeKey = generateKey(translatedName, 'route', context.properties);
+  const component = checkFeatureFlag(page, context, featureFlags);
```
nictownsend:
```suggestion
 const pageEnabled = checkFeatureFlag(context, featureFlags);
```

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -18,44 +22,61 @@ const removeWhitespace = (name) => name.replace(/\s/g, '');
...
-    });
-  } else {
-    leaves.push({
-      parent: `/${extractTopLevelPath(context.path)}`,
-      child: {
-        path: context.path,
-        name: translatedName,
-      },
-    });
+  if (component === page.contentComponent) {
```
nictownsend:
```suggestion
  if (pageEnabled) {
```

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -18,44 +22,61 @@ const removeWhitespace = (name) => name.replace(/\s/g, '');
...
+      icon: context.icon,
+      leaves: [],
+    };
   }
 
   acc.routes.push({
     path: context.path,
     key: routeKey,
-    componentForRoute: page.contentComponent,
+    componentForRoute: component,
```
nictownsend:
```suggestion
    componentForRoute: pageEnabled ? page.component : Error404,
```

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -18,44 +22,61 @@ const removeWhitespace = (name) => name.replace(/\s/g, '');
...
-    });
+  if (component === page.contentComponent) {
+    if (isTopLevelNav(context.path)) {
+      acc.links.push({
+        to: context.path,
+        key: linkKey,
+        children: translatedName,
+      });
+    } else {
+      leaves.push({
```
nictownsend: Why can you not add leaves into the meta at the time? Instead of having to pass in `leaves` to populate and reduce later? It doesn't feel like it's a true reducer.

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
```diff
@@ -67,29 +88,43 @@ const mergeReduce = (acc, cur) =>
...
+            routes: [],
+            meta: {},
+            isComplete,
+          }
+        );
+      })
+      .reduce(mergeReduce, {} as RouterConfig);
+
+    return leaves.reduce((acc, leaf) => {
+      if (acc.meta[leaf.parent]) {
```
nictownsend: Same principle re: reduce - it's more of a map/foreach if we can't do it inside the previous reduce.

nictownsend (changes requested):

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
JQrdan: Solid suggestion, will definitely do this

client/Bootstrap/Navigation/useRouteConfig/useRouteConfig.hook.ts
JQrdan: So this was because there's no guarantee that a parent existed due to page order (`/topics/:name/edit` coming before `/topics`). However, I suppose this could check if a parent is there, and if not, create one in preparation.

nictownsend (approved):

nictownsend (approved):
LGTM - cleaner approach to reducing too is nice

matthew-chirgwin (approved):
Pending @nictownsend 's review, this LGTM

scholzj (approved):

nictownsend (approved):

nictownsend:
> Pending @nictownsend 's review, this LGTM

Sorry - clicked comment instead of approve!

PR #128, opened 2020-11-23
feat: local authentication

### LINKED ISSUE #106: Introduce Passport.js with local strategy for authentication
Add passport as authentication layer onto Express server.

Use a local authentication strategy

Redirect client to `/login` if authentication missing on a protected route

`/login` express endpoint for performing the calls to `mockapi`

nictownsend:
- introduce passport
- add a passport local strategy for authentication
- introduce Authentication interface to contain the multiple
auth checks (authenticate, checkAuth, logout)

Contributes to: strimzi/strimzi-ui#106


nictownsend:
Work to be done:

- [x] Further integration tests
- [x] MockAPI for `/login` purpose

linting/eslint.config.js
```diff
@@ -12,7 +12,7 @@ const rulesets = [
 
 const customRules = {
   // prefer spaces (2) over tabs for indentation - tab width can be changed, space cannot
-  indent: ['error', 2],
+  indent: ['error', 2, { SwitchCase: 1 }],
```
matthew-chirgwin, line 15: What does this new config do (and should the comment be updated to match)?

server/client/client.feature
```diff
@@ -5,44 +5,81 @@ Feature: client module
...
+        And There are files to serve
+        And authentication type 'none' is required
+        And I run an instance of the Strimzi-UI server
+        When I make a 'get' request to '/index.html'
+        Then the file is returned as with the expected configuration included
+
+    Scenario Outline: If assets can be served without authentication, the client module returns the appropriate <StatusCode> return code for a request of <Asset>
+        Given a server with a 'client' configuration
+        And There are files to serve
+        And authentication type 'none' is required
```
matthew-chirgwin, line 34: Being picky, behaviourally/grammatically, would it be better to have three seperate cases?

- And no authentication of the user is required
- And authentication via an identity provider 'type' of the user is required
- And username/password authentication of the user is required

server/client/client.feature
```diff
@@ -5,44 +5,81 @@ Feature: client module
...
+        When I make a 'get' request to '<Asset>'
+        Then I get the expected status code '<StatusCode>' response
+        # if the route (not file) is not matched, we redirect to index.html. Hence / and someroute response
+        Examples:
+            | Asset               | StatusCode |
+            | /index.html         | 200        |
+            | /images/picture.svg | 200        |
+            | /doesnotexist.html  | 404        |
+            | /someroute          | 302        |
+            | /protected.html     | 200        |
```
matthew-chirgwin, line 45: This case (correct me if I am wrong) appears to be removed given the other test changes. The sceanrio here checks public assets can be retrieved without auth - but protected ones cannot

server/client/controller.ts
```diff
@@ -25,7 +25,6 @@ const publicFiles = [
   'images/*',
   'fonts/*',
   'favicon.ico',
-  'index.html',
```
matthew-chirgwin, line 28: Just to confirm, index now protected, as express will redirect to login if needed? What if auth is off?

server/placeholderFunctionsToReplace.ts
```diff
@@ -4,26 +4,4 @@
...
-    case 'none':
-      return (req, res, next) => next();
-    case 'scram':
-    case 'oauth':
-      return (req, res) => res.sendStatus(511); // if auth on, reject for sake of example. This is a middleware, akin to passport doing its checks.
-  }
-};
-
-export { authFunction };
+export {};
```
matthew-chirgwin: Should this not be deleted if nothing uses it?

server/security/router.ts
```diff
@@ -0,0 +1,40 @@
...
+    const routerForModule = express.Router();
+
+    switch (authentication.type) {
+      case authenticationStrategies.SCRAM: {
+        logger.info('Mouting SCRAM security routes');
+        routerForModule.post(scram.login, auth.authenticate, (_req, res) =>
+          res.send(200)
+        );
+        routerForModule.get(scram.login, (_req, res) =>
+          res.send('This will later be the login page')
```
matthew-chirgwin: comment to issue?

server/security/security.feature
```diff
@@ -0,0 +1,62 @@
+# Copyright Strimzi authors.
+# License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+Feature: Security module
+
+  Scenario: SCRAM - authenticate valid credentials
+
+    Given a server with a 'security' configuration
+  # And authentication type 'scram' is required
```
matthew-chirgwin: Why are these disabled? Follow on work?

server/security/security.steps.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { Fusion } from 'jest-cucumber-fusion';
+import 'test_common/commonServerSteps';
```
matthew-chirgwin, line 6: this should not be needed - it should be provided for you

server/security/strategy/strategyFactory.feature
```diff
@@ -0,0 +1,13 @@
...
+# License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+Feature: Strategy Factory
+
+  Scenario Outline: Strategies are returned of the expected type
+
+    When a strategy factory is asked for type '<type>'
+    Then the returned strategy is of type '<expected>'
+    Examples:
+      | type    | expected |
+      | scram   | scram    |
```
matthew-chirgwin, line 11: oauth?

matthew-chirgwin (changes requested):

linting/eslint.config.js
nictownsend: It was prettier and eslint disagreeing over `case` statement indenting; this just fixes that to state the amount of indenting expected.

server/client/controller.ts
nictownsend: Yes, but if auth is off then `checkAuth` on protected routes is a noop

server/placeholderFunctionsToReplace.ts
nictownsend: Left as paradigm seemed useful for future, happy to remove.

server/security/security.feature
nictownsend: Mistake, will add back :)

server/security/security.steps.ts
nictownsend: I thought as much, but it doesn't - https://github.com/b-yond-infinite-network/jest-cucumber-fusion/blob/master/docs/ReusingStepDefinitions.md suggests the same

server/security/strategy/strategyFactory.feature
nictownsend: To be added in the oauth PR :)

server/client/client.feature
nictownsend: I thought I had different scenarios to cover that below?

server/client/client.feature
nictownsend: I'm not sure I understand - how can protected files be blocked if there is no auth mechanism?

nictownsend: I though the "no auth" case showed everything, but when "auth" is enabled, the protected routes required you to have logged in?

nictownsend: Figured out what you mean (files that need auth cannot be retrieved when you're not authenticated) - refactoring now

server/client/client.feature
matthew-chirgwin: Sorry, I was more refering to the definition of the step - does a user care authentication type is none, or that they are/have been authenticated

server/placeholderFunctionsToReplace.ts
matthew-chirgwin: My preference would be that it is - as it has now been superceeded

server/security/security.steps.ts
matthew-chirgwin: I am very curious how any of the other tests work in this case. These steps should be preloaded for everyone here: https://github.com/strimzi/strimzi-ui/blob/ae571ee62875127d08716a42beb1ba9f58f79651/test_common/jest_cucumber_support/index.ts / https://github.com/strimzi/strimzi-ui/blob/74f9304bfd32ad7008c3f6eefd2bef8907a5d2cc/test_common/jest.common.config.js . If you want/need to interact with the world then it make sense to import the functions to be able to do that.

utils/dev_config/server.dev.config.js
```diff
@@ -23,16 +23,20 @@ module.exports = {
   },
   modules: {
     api: true,
-    client: false,
+    client: true,
```
matthew-chirgwin: When developing the UI (which this config is intended for), webpack hosts client files. As a result, we disable the module that would host the files here, to avoid confusion (why are my changes not present etc). What is this being enabled for?

nictownsend: Artifact of testing, removing now.

server/security/security.steps.ts
nictownsend: Raised issue #146 - the existing tests work because they import the world and thus trigger the embedded given/when/then step definitions

server/client/client.feature
nictownsend: My opinion is that in these steps you're a "user" of the server. So you care around the config you have used to create the server.

pmuir (changes requested):
There seem to be some unrelated changes caught up in this PR. Please can you remove them and submit them separately.

nictownsend:
@pmuir sorted - they were whitespace changed from linting whole repo instead of staged files

PR #132, opened 2020-11-27
Adds module federation on top of webpack 5

pmuir:


build/webpack.common.js
```diff
@@ -58,6 +59,29 @@ const withWebpackBundleAnalyzerPlugin = returnPluginWithConfig(
 
 const withTsconfigPathsPlugin = returnPluginWithConfig(TsconfigPathsPlugin, {});
 
+const withModuleFederationPlugin = new webpack.container.ModuleFederationPlugin(
+  {
+    name: federatedModuleName,
+    filename: 'remoteEntry.js',
+    exposes: {
+      './Home': './client/Panels/Home/Home',
```
nictownsend, line 67: What does this mean for each component; is it manually generated? Or can be automated?

dlabaj (approved):

build/webpack.common.js
pmuir: components are manually exported. This is necessary as you don't want to export *everything*

PR #133, opened 2020-11-30
fix: Sort express mountpoints to avoid requests getting hijacked

ajborley:
- This commit reverse-sorts the mountpoints added to express so that
shorter mountpoints such as '/' do not hijack traffic meant for other
mountpoints e.g. '/api' or '/log'. This is not seen when running the
dev server, as the dev server config does not enable the `client`
module.

Contributes to: #42

### DIFF
--- a/server/core/app.ts
+++ b/server/core/app.ts
@@ -75,8 +75,21 @@ export const returnExpress: (
     next();
   });
 
-  Object.entries(routingTable).forEach(
-    ([moduleName, { mountPoint, routerForModule }]) =>
+  Object.entries(routingTable)
+    .map(([moduleName, { mountPoint, routerForModule }]) => ({
+      moduleName,
+      mountPoint,
+      routerForModule,
+    }))
+    .sort(({ mountPoint: mountPointA }, { mountPoint: mountPointB }) => {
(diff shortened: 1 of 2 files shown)


matthew-chirgwin:
Hi @ajborley - am I right in thinking (given the change made), this is to stop https://github.com/strimzi/strimzi-ui/blob/e4499460c95472438309a6544084879e079ce27e/server/client/router.ts#L48 responding to all requests?

ajborley:
@matthew-chirgwin - Yep, the ordering in https://github.com/strimzi/strimzi-ui/blob/master/server/core/modules.ts meant that the express router was mounting the client module on `/` before the config, log or mockapi modules. That means that calls to GET /log or GET /config were being serviced by the client module.

This PR reverse-sorts the list based on the mountpoint so that the order of mounting is now:
 - `/log` (log module)
 - `/config` (config module)
 - `/api` (api module)
 - `/api` (mockapi module)
 - `/` (client module)

It also means, for example, that if we have a 'topics' module mounted at `/api/topics` in the future, calls to that endpoint will get serviced by the topics module rather than the api module.

matthew-chirgwin (approved):

scholzj (approved):

Issue #70, opened 2020-10-30
Topics (listing) page


matthew-chirgwin:
This issue is to cover the discussion and design of the UX for the Topics listing page.

As a part of the discussion, the defintion of a set of Gherkin based behaviours should be identified, which will be used as the base for acceptance testing. The example given in [the UI proposal](https://github.com/strimzi/proposals/blob/master/011-strimzi-ui.md#topic-list-page) is:

```
Scenario: Seeing all available topics:

Given I am a user of the UI
When I navigate to the topic listing
Then see all topics in my Kafka cluster

Scenario: Viewing a specific topic (when it exists in the cluster):
Given I am a user of the UI
When I navigate to the topic listing
And I filter for topic 'SampleTopic'
Then I see topic 'SampleTopic' in the topic list
And Topic 'SampleTopic' has '3' partitions and '2' replicas

Scenario: Viewing a specific topic (when it does not exist in the cluster):
Given I am a user of the UI
When I navigate to the topic listing
And I filter for topic 'SampleTopic'
Then I am told topic 'SampleTopic' does not exist in the cluster
```

A Carbon mockup of this, (again from the proposal) is as follows:

![image](https://user-images.githubusercontent.com/43138971/97716976-2a40dd80-1abc-11eb-8314-bc7504dca92b.png)

This discussion may also identify other UI components required to enable these flows. These should be discussed/designed in their own issues.

matthew-chirgwin:
One question I would raise for this first stage of the Strimzi UI is the need/requirement of a visual rendering of a Navigation component (the left bar in the screenshot). Functionally the navigation will need to exist, but with one page, is there any merit in working on the visuals for one?

chloepoulterdesign:
To confirm @matthew-chirgwin :
1. Is `Create topic` in-scope?
2. Is `Connect to this cluster` (connection details) in-scope?

I've done a mock up to explore this a bit - early screenshots below (NOT FINAL)
![image](https://user-images.githubusercontent.com/47979346/98810774-981fca00-2417-11eb-9556-0fe653ba52b5.png)

I *think* this covers all the gherkin behaviours. Let me know what you think!

Matt, you'll see it's potentially a little different to what you might expect - apparently this Table-header treatment is more accessible than having a page header followed by a table with no header.

matthew-chirgwin:
> To confirm @matthew-chirgwin :
> 
>     1. Is `Create topic` in-scope?
> 
[...]

PR #136, opened 2020-12-02
feat: Add useTopics model hook and query for topics listing

### LINKED ISSUE #124: Implement model & controller for Topics (listing) page
Implement the model & controller for the Topics (listing) page. #70 contains the designs. See also #123 and #122 for the view implementations.

ajborley:
- This commit adds a new `useTopics` model hook, which makes GraphQL
requests to the server to administer Kafka topics.
 - The hook returns a set of sub-hooks, with just the `useGetTopics`
sub-hook implemented for now. `useGetTopics` returns the list of topics
(including name, number of partitions and number of replicas).

Contributes to: #124


client/Models/README.md
```diff
@@ -0,0 +1,37 @@
+# Models
```
matthew-chirgwin: The model for a Panel/Group is a hook, but really should be a private one - ie `useTopics` (I would suggest a rename to `useTopicModel` to be hyper picky) would be defined in a `Model.ts` file in the Topics Panel (and not exported as it is private/internals). 

As for testing, Panels are meant to be behaviourally tested. Ie, a table with data appears with the expected info. As there is no UI (yet), thus no table, it makes asserting (indirectly) that the model is behaving abit more obtuse. I think what would be sensible is to add a placeholder `View.patternfly.ts`/`View.carbon.ts`, which just stringifies the model output. Then, tests can assert the right data is present (just not in a table), which can then be updated when the component work comes along

client/Models/useTopics/Hook.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { GET_TOPICS } from 'Queries/Topics';
+import { useQuery } from '@apollo/client';
+import { useTopicsType } from './useTopics.types';
+
+export const useTopics = (): useTopicsType => {
+  const useGetTopics = () => useQuery(GET_TOPICS);
```
matthew-chirgwin: I would cross reference the behviours described here: https://github.com/strimzi/strimzi-ui/issues/70#issue-733230202

I think we will need to also have a variable for topic name, which gets debounced etc

matthew-chirgwin: Should also add, the consumer of the model is the view. Within reason, if there are things the model can do to make the view as simple as possible (eg if no data returning an empty array), or obfiscate (loading, error states etc) that would be good too.

matthew-chirgwin (changes requested):
Thanks for this Andy!

The core code looks good - as per the doc however, the model hook should be private, and defined in the 'Topics' panel component (along with tests to match/align). I also think a few capabilities may be need adding to the hook too.

Other than that, this LGTM

client/Models/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+This directory contains custom [React Hooks](https://reactjs.org/docs/hooks-intro.html#motivation). Unlike the [Hooks](../Hooks/README.md), these hooks encapsulate specific business logic for views to utilise.
+
+## Test approach
+
+Elements should be tested in a functional manor. See [Test Driven Development](../../docs/Test.md#style-of-test).
```
nictownsend:
```suggestion
Elements should be tested in a functional manner. See [Test Driven Development](../../docs/Test.md#style-of-test).
```

client/Panels/Home/Home.tsx
```diff
@@ -17,11 +18,19 @@ const Home: FunctionComponent = ({ children }) => {
...
+  const { useGetTopics } = useTopics();
+  const { data, loading } = useGetTopics();
+
   return (
     <div className='home'>
       <img src={image} alt='Strimzi logo' />
       <p>Welcome to the Strimzi UI</p>
       {showVersion && isComplete && <p>{`Version: ${version}`}</p>}
+      {loading || !data ? (
+        <p>Loading...</p>
```
nictownsend: Translation ;)

client/Models/README.md
nictownsend: Just to confirm - models live inside the component that needs them, whereas reusable hooks go in the `hooks` dir?

Therefore, `useTopics` is a particular hook (if we want to encapsulate for reuse) - and the `Model` can either call `useTopics` or just the default `useQuery`?

matthew-chirgwin: > Just to confirm - models live inside the component that needs them, whereas reusable hooks go in the `hooks` dir?
> 
> Therefore, `useTopics` is a particular hook (if we want to encapsulate for reuse) - and the `Model` can either call `useTopics` or just the default `useQuery`?

Yes - absolutley. If the intent was to have a reusable hook representing/abstracting the query, than that should be in `Hooks`. Given the context (and other changes), I reviewed/interpreted it as a model hook - and that (as is), model hooks are defined with their components, not in their own section (so I am expecting the code to move, and this readme to be removed).

ajborley: Rearranged/refactored the code into a private hook in the Topics panel component.

client/Models/useTopics/Hook.ts
ajborley: Have added client-side filter handling for the topics listing. The mockapi server doesn't support this yet, so it currently doesn't make any difference when running the webpack-dev-server, but you can see the filter input field changes are debounced correctly by watching the network calls..

client/Models/README.md
ajborley: Fixed - along with all the others :)

client/Panels/Home/Home.steps.tsx
```diff
@@ -52,7 +52,7 @@ When('it is rendered with no version', () => {
   showVersionSet = false;
 });
 
-Then('it should display the expected text', () => {
+Then('it should display the expected text', async () => {
```
matthew-chirgwin, line 55: Wont block the PR on it, but was this intended?

client/Panels/Topics/Topics.steps.tsx
```diff
@@ -0,0 +1,87 @@
...
+    )
+  );
+  jest.runAllTimers();
+});
+
+And(/^I filter for topic '(.+)'$/, async (filter) => {
+  fireEvent.change(renderResult.getByPlaceholderText('filter'), {
+    target: { value: filter },
+  });
+  await waitFor(() => {
```
matthew-chirgwin, line 60: My personal preference would be that rather than using mock timers, we use the apollo helpers, eg https://github.com/strimzi/strimzi-ui/blob/a4c0f81b856c372d2e17d7e4a2194803ccd81d9f/client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.spec.tsx#L290

client/Panels/Topics/useTopicsModel.assets.ts
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright Strimzi authors.
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { generateMockDataResponseForGQLRequest } from 'utils/test/withApollo/withApollo.util';
```
matthew-chirgwin, line 5: `import { generateMockDataResponseForGQLRequest } from 'utils/test'` ?

client/Panels/Topics/useTopicsModel.spec.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useTopicsModel } from './Model';
+import {
+  mockGetTopicsRequests,
+  mockGetTopicsResponses,
+} from './useTopicsModel.assets';
+
+describe('`useTopics` hook', () => {
+  describe('`useGetTopics` hook', () => {
+    beforeEach(() => {
+      jest.useFakeTimers();
```
matthew-chirgwin: Same comment here re use of helper functions, and not to control timers

client/Panels/Topics/Model.ts
```diff
@@ -0,0 +1,41 @@
...
+export const useTopicsModel = (): useTopicsModelType => {
+  const useGetTopics = (filter?: string): useGetTopicsResultType => {
+    // Use debounce to ensure char-by-char changes to the filter string do not cause floods of useQuery network calls
+    const debouncedFilter = useDebounce(filter, 500);
+    // Caching in useQuery means that new network calls are only made when the debouncedFilter value changes
+    return useQuery(GET_TOPICS, { variables: { filter: debouncedFilter } });
+  };
+
+  return {
+    useGetTopics,
```
matthew-chirgwin: This is interesting (and not what I expected - not to say that is wrong!). I expected that we would also be owning/exporting the state for the filter - ie its value and function to update it, which in fact is a (lodash) debounced version. When that debounce clears, then  the value is set, firing the query. Does this mean the view is making use of a `useState` call to track the filter value?

Edit: having seen the code below, I see this is the case. My preference would be the Model does all hook work, and that the view imports as few hooks as possible (for easy sharing of the business logic/keeping the view layers thin)

matthew-chirgwin (changes requested):
Couple of minor comments

scholzj (approved):

pmuir (approved):

ajborley:
@matthew-chirgwin  - I have pushed up a new commit based on your suggestions. It seems to work fine in webpack-dev-server and the e2e server, and tests are passing, but I'm getting an eslint warning that I haven't been able to get past so far.

client/Bootstrap/index.tsx
```diff
@@ -8,17 +8,22 @@ import { init } from 'i18n';
...
 
 init(); //Bootstrap i18next support
 ReactDOM.render(
   <ApolloProvider client={apolloClient}>
     <ConfigFeatureFlagProvider>
       <LoggingProvider>
         <FeatureFlag flag={'client.Pages.PlaceholderHome'}>
-          <Home />
+          <Home>
+            <Topics />
```
pmuir, line 25: Needs a router, not hard coding like this.

client/Panels/Topics/Model.ts
```diff
@@ -0,0 +1,38 @@
...
+ * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
+ */
+import { ChangeEvent, useState, useCallback } from 'react';
+import { GET_TOPICS } from 'Queries/Topics';
+import { useQuery } from '@apollo/client';
+import { useTopicsModelType, topicsType } from './useTopicsModel.types';
+import debounce from 'lodash.debounce';
+
+const onChangeEvent = (
+  evt: ChangeEvent<HTMLInputElement>,
```
pmuir, line 12: Don't use html events here, it breaks the MVVC abstraction. Use scalar types or add an abstraction.

client/Panels/Topics/Model.ts
```diff
@@ -0,0 +1,38 @@
...
+  callWithValue: (value: string) => void
+) => callWithValue(evt.target.value);
+
+export const useTopicsModel = (): useTopicsModelType => {
+  const [filter, setTopicsFilter] = useState();
+  const debouncedUpdateTopicsFilter = useCallback(
+    debounce(setTopicsFilter, 500),
+    []
+  );
+  const { data, loading, error } = useQuery<topicsType>(GET_TOPICS, {
```
pmuir, line 22: Types start with capital letters.

client/Panels/Topics/Topics.steps.tsx
pmuir: Agreed.

client/Panels/Topics/useTopicsModel.types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type topicsType = {
+  topics: topicType[];
+};
+
+export type topicType = {
+  name: string;
+  partitions: number;
+  replicas: number;
+};
```
pmuir, line 27: Use the types generated by the Rama's PR here.

client/Queries/Topics/index.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const GET_TOPICS = gql`
+  query Topics($filter: String) {
+    topics(filter: $filter) {
+      name
+      partitions
+      replicas
+    }
+  }
+`;
```
pmuir, line 15: Not valid graphql schema - update to latest schema from admin server.

pmuir (changes requested):

Issue #127, opened 2020-11-23
Implement Patternfly / Carbon view build time switching.


dlabaj:
Carbon and PatternFly build time switching should be implemented to allow the team build either view for strimzi-ui.

matthew-chirgwin:
Isn't this implemented (and integrated)? https://github.com/strimzi/strimzi-ui/blob/4a5d063b0a894541eed60c2de716e17467db5007/build/webpack.common.js#L61

PR #141, opened 2020-12-03
Example of carbon / patternfly view switching

dlabaj:
We plan to ship carbon as the default view layer. However we do need a way to switch between the patternfly view. We should add an example implementation of this to the client. 

Fixes issue #127


client/jest.config.js
```diff
@@ -15,6 +15,7 @@ const config = {
   moduleNameMapper: {
     ...pathsToModuleNameMapper(compilerOptions.paths, { prefix: '<rootDir>/' }),
     ...jestModuleMapper,
+    "\\.(css|less)$": "<rootDir>/../node_modules/@patternfly/react-styles/__mocks__/styleMock.js"
```
nictownsend, line 18: I thought `jestModuleMapper` brought in fake file support - will it work if you augment `/utils/tooling/aliasHelper.js` to include `less` instead of bringing in `styleMock`?

nictownsend (changes requested):
This contains other commits (including the module federation). From what I can see related to  the patternfly switching, just one comment re mocks.

PR #150, opened 2020-12-10
fix: load the api at a base path

pmuir:
Allows loading the API at a different base path


matthew-chirgwin (commented):
Agree that having the base path as a configurable option makes sense. A couple of quick thoughts on the approach, using what we already have:

- Add a `baseUrl` (or similar) configuration value to the configuration - defaulting to slash, but allowing this to be changed if/as when needed
- For the client either;
  - Use the `base` option in the HTMLWebpackPlugin (using the `baseUrl` value), and make all requests in the client relative (so the base value is encorparated/used by the browser)
  - Send the value from configuration in the HTML (akin to the secuirty), and use that value in any/all requests made

Issue #152, opened 2020-12-14
Production build of UI browser bundle is bunding test libraries


nictownsend:
Browser gives error:

`process is not defined` in `220.bundle.js`.

`npx webpack --config build/webpack.client.prod.js --profile --json > generated/stats.json` and `npx whybundled generated/stats.json --directOnly` shows that test library and cypress are bundled - which use `process.env` under the covers (and cause the bundle to fail).

I believe webpack 4 polyfilled process, whereas webpack 5 no longer does.

PR #153, opened 2020-12-14
fix: stop test assets bundled in production

nictownsend:
- restructure ConfigFeatureFlag assets so they are no longer
exporting test code as runtime dependencies

Contributes to: strimzi/strimzi-ui#152

### DIFF
--- a/client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.assets.ts
+++ b/client/Contexts/ConfigFeatureFlag/ConfigFeatureFlag.assets.ts
@@ -2,16 +2,10 @@
  * Copyright Strimzi authors.
  * License: Apache License 2.0 (see the file LICENSE or http://apache.org/licenses/LICENSE-2.0.html).
  */
-// deliberately not using the barrel as causes a circular dependency
-import {
-  generateMockDataResponseForGQLRequest,
-  generateMockErrorResponseForGQLRequest,
-} from 'utils/test/withApollo/withApollo.util';
 import {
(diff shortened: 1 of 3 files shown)


nictownsend:
@JQrdan - any better way you'd prefer to split the assets?

matthew-chirgwin (commented):
Agree that these peieces should not be used/exported in production via the barrel.

Out of interest, would a:
```
export {publicItemOne, publicItemTwo} from './ConfigFeatureFlag.assets'
```

replacing https://github.com/strimzi/strimzi-ui/blob/2ee0ba3ce3d76073cc9ba1178e54202fc4828255/client/Contexts/ConfigFeatureFlag/index.ts#L6 achieve the same effect (without the need for extra files)?

nictownsend:
> Agree that these peieces should not be used/exported in production via the barrel.
> 
> Out of interest, would a:
> 
> ```
> export {publicItemOne, publicItemTwo} from './ConfigFeatureFlag.assets'
> ```
> 
> replacing
> 
> https://github.com/strimzi/strimzi-ui/blob/2ee0ba3ce3d76073cc9ba1178e54202fc4828255/client/Contexts/ConfigFeatureFlag/index.ts#L6
> 
> achieve the same effect (without the need for extra files)?

Unfortunately not - it seems like treeshaking isn't smart/configured to notice that the imports in `ConfigFeatureFlag.assets` are not used by the exported functions. Am happy to raise a following issue for backlog to investigate this though.

